PR #4, opened 2021-07-27
Zok to ABY pipeline 

edwjchen:
***Completed***
- Lowering layer from CirC IR to ABY Circuits
- Output layer to write translated code into ABY submodule and update CMake files
- Python testing framework for ABY tests

***Future Action Items***
ABY support for:
  - Power
  - Negation
  - Division
  - Remainder
  - Left shift
  - Right shift
  - Conditionals
  
Tests:
  - Add test cases with public values


src/target/aby/trans.rs
```diff
@@ -0,0 +1,477 @@
...
+                        self.cache.insert(
+                            t.clone(),
+                            EmbeddedTerm::Bv(format!("circ->PutXORGate({}, {})", a, b)),
+                        );
+                    }
+                    BvNaryOp::Or => {
+                        self.cache.insert(
+                            t.clone(),
+                            EmbeddedTerm::Bv(format!(
+                                "((BooleanCircuit *)circ)->PutORGate({}, {})",
```
alex-ozdemir, line 374: I'm intrigued by this cast.

src/target/aby/trans.rs
```diff
@@ -0,0 +1,477 @@
...
+                self.cache.insert(
+                    t.clone(),
+                    EmbeddedTerm::Bool(format!("circ->PutMUXGate({}, {}, {})", a, b, sel)),
+                );
+            }
+            Op::BvNaryOp(o) => {
+                let a = self.get_bv(&t.cs[0]);
+                let b = self.get_bv(&t.cs[1]);
+
+                match o {
```
alex-ozdemir, line 363: Save the weird cast, all of these have the same structure, so you could so something like

```
self.cache.insert(
  t.clone(),
  EmbeddedTerm::Bv(format!("circ->{}({}, {})",
    match o {
      BvNaryOp::Xor => "PutXORGate",
      // ...
    }, a, b)));
```

src/target/aby/trans.rs
```diff
@@ -0,0 +1,477 @@
...
+                match op {
+                    BvBinPred::Uge => {
+                        let eq = self.embed_eq(t.clone(), &t.cs[0], &t.cs[1]);
+                        self.cache.insert(
+                            t.clone(),
+                            EmbeddedTerm::Bool(format!(
+                                "((BooleanCircuit *)circ)->PutORGate(circ->PutGTGate({}, {}), {})",
+                                a, b, eq
+                            )),
+                        );
```
alex-ozdemir, line 281: Did you consider embedding this (and ULT) as NOT composed with LT (and similar for ULT)? I worry that both EQ and LT/GT are expensive in MPC, while NOT is cheap.

src/target/aby/trans.rs
```diff
@@ -0,0 +1,477 @@
...
+                let a = self.get_bv(&t.cs[1]).clone();
+                let b = self.get_bv(&t.cs[2]).clone();
+                self.cache.insert(
+                    t.clone(),
+                    EmbeddedTerm::Bool(format!("circ->PutMUXGate({}, {}, {})", a, b, sel)),
+                );
+            }
+            Op::BvNaryOp(o) => {
+                let a = self.get_bv(&t.cs[0]);
+                let b = self.get_bv(&t.cs[1]);
```
alex-ozdemir, line 361: This isn't quite right. While your tests seem to only be generating _binary_ additions, multiplications, etc., these are n-ary operators, and the compiler might give you n-ary instances of them. Perhaps you need to fold over the list here?

alex-ozdemir: You should be able to create a problematic test case by just adding 3 numbers (unless I'm missing something)

alex-ozdemir: Alternatively, you might write and IR-to-IR transformation pass that just binarizes the operator. That would be more re-usable, but perhaps a bit more code.

src/target/aby/trans.rs
```diff
@@ -0,0 +1,477 @@
...
+                        self.cache.insert(
+                            t.clone(),
+                            EmbeddedTerm::Bv(format!("circ->PutMULGate({}, {})", a, b)),
+                        );
+                    }
+                }
+            }
+            Op::BvBinOp(o) => {
+                let a = self.get_bv(&t.cs[0]);
+                let b = self.get_bv(&t.cs[1]);
```
alex-ozdemir, line 401: This is good. These ops are binary.

src/target/aby/trans.rs
```diff
@@ -0,0 +1,477 @@
...
+            Op::Not => {
+                let a = self.get_bool(&t.cs[0]);
+                self.cache.insert(
+                    t.clone(),
+                    EmbeddedTerm::Bool(format!("((BooleanCircuit *)circ)->PutINVGate({})", a)),
+                );
+            }
+            Op::BoolNaryOp(o) => {
+                let a = self.get_bool(&t.cs[0]).clone();
+                let b = self.get_bool(&t.cs[1]).clone();
```
alex-ozdemir, line 243: See comment below about n-ary bitvector operators.

alex-ozdemir (approved):
Thanks for doing this Edward, and good work!

I left a few comments about style and correctness. You can address them in this PR, or defer them---your call.

src/target/aby/trans.rs
edwjchen: Similar with the negation gate, OR apparently only works with Boolean Circuits... 
[Searching for PutORGate only shows boolean_circuits.cpp](https://github.com/encryptogroup/ABY/search?q=putorgate)

alex-ozdemir: Ah, makes sense. I thought this meant something else, which is why I was confused.

src/target/aby/trans.rs
edwjchen: I added an nary_add test case:

def main(private<1> u32 a, private<2> u32 b,  private<2> u32 c) -> u32:
  return a + b + c

and I got back this set of terms:

(bvadd (bvadd main_f0_lex0_a_v0 main_f0_lex0_b_v0) main_f0_lex0_c_v0)

I checked the MPC optimizations and it looks like we removed the flatten optimization since all of the gates only take in binary inputs anyways. 

Without the flatten optimization, I think there shouldn't be any nary-terms (unless I'm also missing something here).

alex-ozdemir: Ah, I see---I was missing something. We're getting lucky here, in a sense, because the ZoKrates front-end is not producing an n-ary term for `a + b + c`.

That doesn't mean the front-end *never* produces n-ary terms though. I bet it does in some array cases.

src/target/aby/trans.rs
edwjchen: very cool!

src/target/aby/trans.rs
edwjchen: Will punt this to the next PR!

src/target/aby/trans.rs
edwjchen: I see, is there any difference between these two options other than reusability?
- writing an IR-to-IR transformation that binarizes the operator 
- using a for loop to iterate and append a gate for each of the inputs

alex-ozdemir: I don't see any.

edwjchen: I'll add the binarize translation layer into the IR in the next PR!

alex-ozdemir:
Woo! :tada:

PR #22, opened 2021-10-20
C Frontend

edwjchen:
Working C frontend supporting:
- one main function
- binary arithmetic 
- if-else stmts
- array select and store

Supports both integer and boolean types.


alex-ozdemir (changes requested):
This is pretty close to a state where I'd be happy to merge it. There a few small issues noted inline, that should be fixed.

We may want to fix up the array notion at some point, but this is fine for now.

examples/C/mpc/unit_tests/array_tests/2pc_array_sum.c
```diff
@@ -0,0 +1,5 @@
+int main(__attribute__((private(0))) int a, __attribute__((private(1))) int b) { 
+    int c[2][5][6];
```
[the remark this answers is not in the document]
edwjchen: ah! I think I was using this test to see how multiple dimensions arrays were initialized and I didn't remove `c`. Thanks for the catch!

src/front/c/mod.rs
```diff
@@ -0,0 +1,438 @@
...
+        match c {
+            Constant::Integer(i) => CTerm {
+                term: CTermData::CInt(32, bv_lit(i.number.parse::<i32>().unwrap(), 32)),
+                udef: false,
+            },
+            _ => unimplemented!("Constant {:#?} hasn't been implemented", c),
+        }
+    }
+
+    fn get_bin_op(&self, op: BinaryOperator) -> fn(CTerm, CTerm) -> Result<CTerm, String> {
```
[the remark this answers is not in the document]
edwjchen: Haha :) I think you were the one who wrote this part first! Was very cool getting to read something like this!

src/front/c/mod.rs
```diff
@@ -0,0 +1,438 @@
...
+            _ => unimplemented!("Invalid left hand value"),
+        }
+    }
+
+    /// Interpret the party association of input parameters 
+    fn interpret_visibility(&mut self, ext: &DeclarationSpecifier) -> Option<PartyId> {
+        if let DeclarationSpecifier::Extension(nodes) = ext {
+            assert!(nodes.len() == 1);
+            let node = nodes.first().unwrap();
+            if let Extension::Attribute(attr) = &node.node {
```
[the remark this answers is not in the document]
edwjchen: Hmm which parts exactly? It seems a little tricky with the call to self.gen_expr in the middle.

src/front/c/mod.rs
```diff
@@ -0,0 +1,668 @@
...
+                for (i, v) in values.iter().enumerate() {
+                    let offset = bv_lit(i, 32);
+                    let v_ = v.term.term(&self.circ);
+                    self.circ.store(id, offset, v_);
+                }
+
+                CTerm {
+                    term: CTermData::CArray(inner_type, Some(id)),
+                    udef: false,
+                }
```
alex-ozdemir, line 413: This is a very common pattern. I'd consider adding a one-argument constructor.

src/front/c/mod.rs
```diff
@@ -0,0 +1,668 @@
...
+        let res = self.circ.declare_init(
+            decl_info.name,
+            derived_ty.clone(),
+            Val::Term(cast(Some(derived_ty), expr.clone())),
+        );
+        self.unwrap(res);
+        expr
+    }
+
+    fn get_const_iters(&mut self, for_stmt: ForStatement) -> Option<ConstIteration> {
```
alex-ozdemir, line 451: This function isn't quite right b/c the loop body could modify the iteration variable. We had a nice way to detect that in the Haskell compiler that would be a bit harder to replicate here. For now, why don't you write a comment noting the inaccuracy.

src/front/c/mod.rs
```diff
@@ -0,0 +1,668 @@
...
+        let incr = incr_.val;
+
+        // TODO: fix naming
+        if cond_name.is_empty() {
+            cond_name = init_name.clone();
+        }
+
+        if init_name == cond_name && cond_name == incr_name {
+            return Some(ConstIteration {
+                name: init_name,
```
alex-ozdemir, line 535: Is this field dead in the callee?

src/front/c/mod.rs
```diff
@@ -0,0 +1,668 @@
...
+                                    .borrow_mut()
+                                    .outputs
+                                    .extend(ret_terms);
+                            }
+                            Mode::Proof => {
+                                let ty = fn_info.ret_ty.as_ref().unwrap();
+                                let name = "return".to_owned();
+                                let term = r.unwrap_term();
+                                let _r = self.circ.declare(name.clone(), ty, false, PROVER_VIS);
+                                self.circ.assign_with_assertions(name, term, ty, PUBLIC_VIS);
```
alex-ozdemir, line 658: I would actually do `unimplemented!()` here, unless you're going to test this support.

src/front/c/term.rs
```diff
@@ -0,0 +1,583 @@
...
+    } else if is_signed_int(a_prom_ty.clone()) == is_signed_int(b_prom_ty.clone()) {
+        if int_conversion_rank(a_prom_ty.clone()) < int_conversion_rank(b_prom_ty.clone()) {
+            return (cast(Some(b_prom_ty), a_prom), b_prom);
+        } else {
+            return (a_prom, cast(Some(a_prom_ty), b_prom));
+        }
+    } else {
+    }
+
+    unimplemented!("Not implemented case in iUAC");
```
alex-ozdemir, line 157: I'd move this into the `else`.

src/front/zokrates/mod.rs
```diff
@@ -611,6 +585,8 @@ impl<'ast> ZGen<'ast> {
         for (p, f) in &t {
             self.file_stack.push(p.to_owned());
             for func in &f.functions {
+                println!("fn {} in {}", func.id.value, self.cur_path().display());
+
```
alex-ozdemir, line 589: Remove this? Also, see [here](https://docs.rs/env_logger/latest/env_logger/index.html) for how to use the logging infrastructure. Quickstart: set `RUST_LOG=debug ` before running the binary.

third_party/ABY_templates/cpp_template.txt
```diff
@@ -22,6 +22,5 @@ int32_t test_{fn}_circuit(std::map<std::string, std::string> params, e_role role
 
 	party->ExecCircuit();
 	flush_output_queue(out_q, role, bitlen);
-	delete party;
```
alex-ozdemir, line 25: Why this change?

src/ir/opt/binarize.rs
```diff
@@ -0,0 +1,120 @@
...
+
+fn binarize(op: &Op, children: &[Term]) -> Term {
+    if children.is_empty() || children.len() == 1 {
+        term(op.clone(), children.to_vec())
+    } else if children.len() == 2 {
+        return term![op.clone(); children[0].clone(), children[1].clone()];
+    } else {
+        return term![op.clone(); children[0].clone(), binarize(op, &children[1..].to_vec())];
+    }
+}
```
alex-ozdemir, line 24: I'd recommend using `Iterator::fold` instead of recursion here. We build some *very* large n-ary terms, and don't want to blow the stack.

src/ir/opt/cfold.rs
```diff
@@ -85,6 +85,8 @@ pub fn fold_cache(node: &Term, cache: &mut TermMap<Term>) -> Term {
                 match (o, c0.as_bv_opt(), c1.as_bv_opt()) {
                     (Sub, Some(a), Some(b)) => cbv(a.clone() - b.clone()),
                     (Sub, _, Some(b)) if b.uint() == &Integer::from(0) => Some(c0.clone()),
+                    // TODO: is this right?
+                    (Udiv, _, Some(b)) if b.uint() == &Integer::from(0) => Some(c0.clone()),
```
alex-ozdemir, line 89: Not quite. We're following SMT-LIB, which says this should be the maximum representable value:
> [[(bvudiv s t)]] := if bv2nat([[t]]) = 0
                       then λx:[0, m). 1
                       else nat2bv[m](bv2nat([[s]]) div bv2nat([[t]]))

See: https://smtlib.cs.uiowa.edu/theories-FixedSizeBitVectors.shtml or https://github.com/circify/circ/blob/master/src/ir/term/bv.rs#L38-L61

src/front/c/mod.rs
```diff
@@ -0,0 +1,668 @@
...
+    fn gen(&mut self) {
+        let TranslationUnit(nodes) = self.tu.clone();
+        for n in nodes.iter() {
+            match n.node {
+                ExternalDeclaration::Declaration(ref decl) => {
+                    debug!("{:#?}", decl);
+                }
+                ExternalDeclaration::FunctionDefinition(ref fn_def) => {
+                    debug!("{:#?}", fn_def.node.clone());
+                    // println!("function: {:#?}", fn_def.node);
```
alex-ozdemir, line 625: rm

alex-ozdemir (changes requested):
I looked at the C FE and your IR/circify changes quite closely. I only skimmed the ABY BE changes. I left a few comments inline.

One other question: 
* what is the `.pyc` file?

edwjchen:
> * what is the `.pyc` file?

the `.pyc` was just a cache bytecode of the python test script. I adjusted the `make clean` command to remove them!

alex-ozdemir (approved):
Great!

PR #33, opened 2021-12-26
IR-based Zokrates front-end

alex-ozdemir:
The ZoKrates front-end now represents ZoK arrays as IR arrays, and ZoK structures as (type-tagged) IR tuples.

During this change, I discovered that IR support for eliminating tuples and arrays was not complete.

Thus the change list is:
1. The ZoK front-end uses IR arrays and tuples
2. Improve IR passes for array and tuple elimination
3. Enforce `cargo fmt` in CI
4. Bugfix: handle ZoK accessors in L-values in the correct order
5. Bugfix: add array evaluation to the IR

This PR **does not**:
* implement an array flattening pass
* implement permutation-based memory-checking

Benefits:
* The ZoK->R1CS compiler is now ~5.88x faster (as defined by the time it takes to run the tests in `master`'s `scripts/zokrates_test.zsh` script: this goes from `8.59s` to `1.46s`)
   * For benchmarks with multi-dimensional arrays, the ZoK->R1CS compiler can now compile them with reasonable speed. Before it it would time out on even tiny examples.
* The ZoK->R1CS compiler will be able to benefit from future memory-checking improvements
* IR support for arrays and tuples is complete now, making those parts of the IR more accessible to future front-ends.


src/front/zokrates/term.rs
```diff
@@ -19,24 +19,37 @@ lazy_static! {
...
     Uint(usize),
     Bool,
     Field,
-    Struct(String, BTreeMap<String, Ty>),
+    Struct(String, Vec<(String, Ty)>),
     Array(usize, Box<Ty>),
 }
 
+fn lookup<'a>(map: &'a Vec<(String, Ty)>, key: &str) -> Option<(usize, &'a Ty)> {
+    let idx = map.binary_search_by_key(&key, |p| p.0.as_str()).ok()?;
```
kwantam, line 36: So I guess this means there's an invariant that map is sorted by String, right? Should this be documented in the definition of `Ty`? Are there `debug_assert`s that would help with this? (Is this paranoia?)

kwantam: Ah, I see, `new_struct` ensures this. Would it perhaps make sense to make the `Vec<(String, Ty)>` a private wrapper type that can *only* be constructed with a function that enforces the invariant? Because as it stands now, I can construct a Struct that is broken, no?

kwantam (approved):
Wow, nice!

Integrating this with #30 will be fun... :)

examples/circ.rs
```diff
@@ -177,22 +177,28 @@ fn main() {
...
                 Opt::Flatten,
                 //Opt::FlattenAssertions,
                 Opt::Inline,
-                Opt::Mem,
+                Opt::Tuple,
+                Opt::ConstantFold,
+                Opt::Obliv,
+                Opt::Tuple,
+                Opt::LinearScan,
+                Opt::Tuple,
```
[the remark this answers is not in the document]
alex-ozdemir: Yes; I've left a comment to this effect.

examples/circ.rs
```diff
@@ -215,12 +221,12 @@ fn main() {
             let r1cs = to_r1cs(cs, circ::front::zokrates::ZOKRATES_MODULUS.clone());
             println!("Pre-opt R1cs size: {}", r1cs.constraints().len());
             let r1cs = reduce_linearities(r1cs);
+            println!("Final R1cs size: {}", r1cs.constraints().len());
             match action {
-                ProofAction::Count => {
-                    println!("Final R1cs size: {}", r1cs.constraints().len());
-                }
+                ProofAction::Count => {}
```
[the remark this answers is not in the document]
alex-ozdemir: Sounds good to me.

src/front/zokrates/term.rs
alex-ozdemir: Yeah, I was being a little lazy here. I've split it out into its own structure and module.

kwantam (approved):
Looks great!

PR #55, opened 2022-02-27
Added Map to IR (evaluation and typechecking)

Northrim:


src/ir/term/mod.rs
```diff
@@ -1252,175 +1258,225 @@ impl Value {
...
-                let c1 = vs.get(&c.cs[1]).unwrap().as_bool() as u8;
-                let c2 = vs.get(&c.cs[2]).unwrap().as_bool() as u8;
-                Value::Bool(c0 + c1 + c2 > 1)
+        eval_value(vs, h, c.clone());
+    }
+    vs.get(t).unwrap().clone()
+}
+
+/// Helper function for eval function. Handles a single term
+fn eval_value(vs: &mut HConMap<HConsed<TermData>, Value>,
```
edwjchen: Can  we use `vs: &mut TermMap<Value>` here?

src/ir/term/mod.rs
```diff
@@ -1252,175 +1258,225 @@ impl Value {
...
-                let c2 = vs.get(&c.cs[2]).unwrap().as_bool() as u8;
-                Value::Bool(c0 + c1 + c2 > 1)
+        eval_value(vs, h, c.clone());
+    }
+    vs.get(t).unwrap().clone()
+}
+
+/// Helper function for eval function. Handles a single term
+fn eval_value(vs: &mut HConMap<HConsed<TermData>, Value>, 
+                h: &FxHashMap<String, Value>, c: HConsed<TermData>) -> Value{
```
edwjchen: Can we use `c: Term` here?

src/ir/term/mod.rs
```diff
@@ -1252,175 +1258,225 @@ impl Value {
...
+        Op::Select => {
+            let a = vs.get(&c.cs[0]).unwrap().as_array().clone();
+            let i = vs.get(&c.cs[1]).unwrap();
+            a.clone().select(i)
+        },
+        Op::Map(op) => {
+            let arg_cnt = c.cs.len();
+            let arr_size = vs.get(&c.cs[0]).unwrap().as_array().size;
+            let mut term_vecs = vec![Vec::new(); arr_size];
+            //2D vector: term_vecs[i] will store a vector of all the i-th index
```
edwjchen: nit: we can move the comments above `let mut term_vecs = ...`

src/ir/term/mod.rs
```diff
@@ -1252,175 +1258,225 @@ impl Value {
...
+            a.clone().select(i)
+        },
+        Op::Map(op) => {
+            let arg_cnt = c.cs.len();
+            let arr_size = vs.get(&c.cs[0]).unwrap().as_array().size;
+            let mut term_vecs = vec![Vec::new(); arr_size];
+            //2D vector: term_vecs[i] will store a vector of all the i-th index 
+            //  entries of the array arguments
+
+            //Value::BitVector(BitVector::new(uint.into(),width,))
```
edwjchen, line 1440: nit: can remove this comment!

src/ir/term/mod.rs
```diff
@@ -1252,175 +1258,225 @@ impl Value {
...
-                let i = vs.get(&c.cs[1]).unwrap().clone();
-                let v = vs.get(&c.cs[2]).unwrap().clone();
-                Value::Array(a.clone().store(i, v))
             }
-            Op::Select => {
-                let a = vs.get(&c.cs[0]).unwrap().as_array().clone();
-                let i = vs.get(&c.cs[1]).unwrap();
-                a.clone().select(i)
+            //Cloning the first arg just for formatting
+            let mut res = vs.get(&c.cs[0]).unwrap().as_array().clone();
```
edwjchen, line 1450: We could create a new array term by using: `Array::default()` from src/ir/term/mod.rs. We would just need to get the key sort, value sort, and size from vs. 

@alex-ozdemir thoughts on this? Since we are already replacing all of the values in the array, either approach is logically the same.

src/ir/term/mod.rs
```diff
@@ -1252,175 +1258,225 @@ impl Value {
...
-    }
-    vs.get(t).unwrap().clone()
+            Value::Array(res)
+        }
+        o => unimplemented!("eval: {:?}", o),
+    };
+    vs.insert(c, v.clone());
+
+    v
+    //println!("Eval {}\nAs   {}", c, v);
```
edwjchen, line 1464: nit: can remove this comment or change to `debug!("Eval {}\nAs {}", c, v);`!

src/ir/term/mod.rs
```diff
@@ -1252,175 +1258,225 @@ impl Value {
...
+}
+
+#[macro_export]
```
edwjchen, line 1467: I think we can keep this macro in the same spot as before below,

```
/// Make a bit-vector constant term.
pub fn bool_lit(b: bool) -> Term {
    leaf_term(Op::Const(Value::Bool(b)))
}
```

src/ir/term/test.rs
```diff
@@ -14,6 +15,37 @@ fn eq() {
...
+        vec![bv(0b0001, 4), bv(0b0010, 4), bv(0b0011, 4), bv(0b0100, 4)]);
+
+    let a2 = 
+        make_array(Sort::BitVector(32), Sort::BitVector(4),
+        vec![bv(0b0001, 4), bv(0b0100, 4), bv(0b1001, 4), bv(0b0000, 4)]);
+
+    let t_eq = term![Op::Map(Box::new(Op::Eq)); a1.clone(), a2.clone()];
+    let t_add = term![Op::Map(Box::new(BV_ADD)); a1.clone(), a2.clone()];
+
+    println!("MAP EQ  Value is: {}", eval(&t_eq, &FxHashMap::default()));
```
edwjchen, line 50: let's assert the value of the `eval` results instead of printing them out!

src/ir/term/test.rs
```diff
@@ -14,6 +15,37 @@ fn eq() {
...
+    let a1 = 
+        make_array(Sort::BitVector(32), Sort::Bool,
+            vec![bool(true), bool(true), bool(false), bool(false)]);
+    let a2 = 
+        make_array(Sort::BitVector(32), Sort::Bool,
+            vec![bool(true), bool(false), bool(true), bool(false)]);
+    let t = term![Op::Map(Box::new(Op::Eq)); a1, a2];
+
+    let val = eval(&t, &FxHashMap::default());
+    println!("Value is: {}", val);
```
edwjchen, line 33: change to `assert!` on an array term with value: `vec![bool(true), bool(false), bool(false), bool(false)]`

src/ir/term/test.rs
```diff
@@ -14,6 +15,37 @@ fn eq() {
...
+
+    let a2 = 
+        make_array(Sort::BitVector(32), Sort::BitVector(4),
+        vec![bv(0b0001, 4), bv(0b0100, 4), bv(0b1001, 4), bv(0b0000, 4)]);
+
+    let t_eq = term![Op::Map(Box::new(Op::Eq)); a1.clone(), a2.clone()];
+    let t_add = term![Op::Map(Box::new(BV_ADD)); a1.clone(), a2.clone()];
+
+    println!("MAP EQ  Value is: {}", eval(&t_eq, &FxHashMap::default()));
+    println!("MAP ADD Value is: {}", eval(&t_add, &FxHashMap::default()))
```
edwjchen, line 51: same!

src/ir/term/ty.rs
```diff
@@ -128,6 +128,49 @@ pub fn check_raw(t: &Term) -> Result<Sort, TypeError> {
...
+                    key_sort = k.clone();
+                    size = s.clone();
+                }
+                Err(e) => {
+                    error = Some(e);
+                }
+            }
+
+            for i in 0..arg_cnt {
+                match array_or(&check_raw(&t.cs[i])?, "map inputs") {
```
edwjchen, line 150: should we use `arrmap_or` here as well to check that all input array sizes are the same? Then we can also combine both for loops together.

src/ir/term/ty.rs
```diff
@@ -128,6 +128,49 @@ pub fn check_raw(t: &Term) -> Result<Sort, TypeError> {
...
+                    Err(e) => {
+                        error = Some(e);
+                    }
+                }
+            }
+
+            match error {
+                Some(e) => Err(e),
+                None => {
+                    let dummy_term = term((**op).clone(), dterm_cs);
```
edwjchen, line 163: nit: rename `dummy_term` to `term_`

edwjchen (changes requested):

edwjchen:
Make sure to run:
```
python3 driver.py --format
python3 driver.py --lint
python3 driver.py --clean
```
before pushing!

src/ir/term/mod.rs
alex-ozdemir: Either approach is good in my mind.

src/ir/term/mod.rs
```diff
@@ -1252,175 +1258,227 @@ impl Value {
...
-                let a = vs.get(&c.cs[0]).unwrap().as_array().clone();
-                let i = vs.get(&c.cs[1]).unwrap();
-                a.clone().select(i)
+            //Cloning the first arg just for formatting
+            let mut res = vs.get(&c.cs[0]).unwrap().as_array().clone();
+            for i in 0..arr_size {
+                let t = term((**op).clone(), term_vecs[i].clone());
+                let val = eval_value(vs, h, t);
+                res.map
+                    .insert(Value::BitVector(BitVector::new(Integer::from(i), 32)), val);
```
alex-ozdemir, line 1455: I'd like to avoid assuming that the array is indexed by `(bv 32)`. I think we should accept
any BV and prime-field element, and any boolean.

src/ir/term/mod.rs
```diff
@@ -1252,175 +1258,227 @@ impl Value {
...
+            Value::Array(a.clone().store(i, v))
+        }
+        Op::Select => {
+            let a = vs.get(&c.cs[0]).unwrap().as_array().clone();
+            let i = vs.get(&c.cs[1]).unwrap();
+            a.clone().select(i)
+        }
+        Op::Map(op) => {
+            let arg_cnt = c.cs.len();
+            let arr_size = vs.get(&c.cs[0]).unwrap().as_array().size;
```
alex-ozdemir, line 1435: I think that you should explicitly check the key sort so that you can generate a good error message. See comment below for permitted key sorts.

src/ir/term/mod.rs
Northrim: I've though about this more and it looks like I should use Array::default() instead of cloning. If I use the cloning approach, it causes the output array term to have an incorrect value type in some cases.

Northrim: I'm trying to use the check() function from src/ir/term/ty.rs to figure out the type to use for Array::default(), but I am running into an issue where check(&c) is returning Sort::bool when "c" is a Map term.
I looked a bit further into this bug and it looks like the type of the Map term is stored as a Boolean in TREM_TYPES, so the check_raw() is returning in its first 3 lines:
```
if let Some(s) = TERM_TYPES.read().unwrap().get(&t.to_weak()) {
        return Ok(s.clone());
    }
```
Is there a way I can fix this issue?

src/ir/term/mod.rs
Northrim: When the indices are bitvectors, it is simple to iterate through the terms, as I can use a for loop.
How should I iterate through the indices for these other index types?

alex-ozdemir: You can do a similar thing for boolean and field elements: https://github.com/circify/circ/blob/master/src/ir/term/mod.rs#L835-L838

src/ir/term/mod.rs
alex-ozdemir: `TERM_TYPES` is just a cache for `check` (and friends), so the relevant question is when we put the wrong sort into the map. I'd just print out every store to the map, and search for the bad insertion.

src/ir/term/mod.rs
Northrim: Just to clarify, should I case on the index type and write separate iterating code for each?

alex-ozdemir: I was going to say that you can just call `elems_iter`, but I now see that doesn't *quite* work.

You should extract a value-based (rather than term-based) version of `elems_iter`, and call the extracted function.

src/ir/term/mod.rs
Northrim: Looks like there was a bug in my check_rec() implementation. I've made the fix and the new approach for initializing the array is working as intended.

src/ir/term/ty.rs
Northrim: I think it would be better if I added the check for the input array sizes to rec_check_raw() instead, since
check_raw() is mostly used to retrieve the type of a term, while rec_check_raw() is used to verify that the terms are well typed.
Do think this would be better?

Also, should I check that the key types are consistent throughout the input arrays as well?

src/ir/term/mod.rs
Northrim: Got it.
When I look at elems_iter, it looks as though it iterates through all of the possible values of the index type.

I am assuming that in my iterator for map, the number of elements I iterate should depend on the size of the arrays I am mapping on. In the case where the key_type is a bitvector or field and the array size is x, can I assume that the indices are going to be 0,1,2,...x-1?

src/ir/term/mod.rs
Northrim: Won't this check be done by the type checking functions?

alex-ozdemir: Good point!

src/ir/term/mod.rs
alex-ozdemir: Yep!

alex-ozdemir (approved):
Looks good.

edwjchen (approved):
lgtm!

PR #65, opened 2022-03-22
Updated ABY VM to include `IN` bytecode instruction

edwjchen:


kwantam:
Something is broken with this PR. `src/front/zokrates` no longer exists, but this PR appears to contain diffs for files in that directory.

edwjchen:
@kwantam Thanks for the catch! I've removed the directory; must have been added back in during the merge process somehow.

src/target/aby/trans.rs
```diff
@@ -97,6 +103,13 @@ impl ToABY {
...
 
+    fn unwrap_vis(&self, name: &str) -> u8 {
+        match self.md.input_vis.get(name).unwrap() {
+            Some(role) => *role,
+            None => PUBLIC,
+        }
+    }
+
     /// Return constant gate evaluating to 1
     // fn one(s_type: &str) -> String {
```
Northrim, line 114: You should delete this function if it is not needed.

src/target/aby/trans.rs
```diff
@@ -246,24 +273,38 @@ impl ToABY {
         let share = self.get_share_name(&t);
         let s = self.term_to_share_cnt.get(&t).unwrap();
         match &t.op {
-            Op::Var(_, Sort::BitVector(_)) => {
+            Op::Var(name, Sort::BitVector(_)) => {
+                if !self.inputs.contains(&t) && self.md.input_vis.contains_key(name) {
+                    let term_name = ToABY::get_var_name(&t);
```
Northrim, line 278: Why is it that for the other helper functions, you use self."function_name", but for get_var_name(), you use ToABY::get_var_name()?

Is it because get_var_name() does not need to take &self as an argument?

Northrim (changes requested):

src/target/aby/trans.rs
edwjchen: `self.get_share_name()` needs to reference `self.term_to_share_cnt`, which requires `self`.

`get_var_name` is getting the string from a `Term`, which doesn't require `self`. 

You can look through the function definitions for a better understanding!

Northrim (approved):
lgtm!

PR #66, opened 2022-03-24
field flexibility ; add limit for linearity reduction

### LINKED ISSUE #31: moduli and types for field elements
As @alex-ozdemir has noted elsewhere, one thing we eventually want to clean up is the way field elements are handled. This thread is intended to kick off a discussion of how we want to do that. Very brief background (I'm sure Alex can add a lot more info here):

Right now, we use `rug::Integer` values for field types. Unfortunately, they are not super fast for modular arithmetic, which is a problem we'd like to solve.

One thing we could do is fix a field-specific type in the IR at (CirC) compile time. That gives us speed, but not convenience. Note, however, that the Z# front-end fixes the modulus at compile time anyway, so it wouldn't materially change things convenience-wise to switch to field-specific types, at least for this front-end (but of course it would be much faster!).

A related option is to make the IR agnostic to field types, by writing down (or borrowing) traits that capture the functionality we need, and having the front-end decide what concrete type implements those traits. (I bet we could just use the `ff` traits for this, and in principle we could wrap `rug::Integer` values in a thin shim to get runtime modulus selection for front-ends that want it.) The negative way of looking at this is that it kicks the can to the front-end. The more positive way to look at it is that it gives front-ends the opportunity to optimize themselves for particular fields! I'd bet that in many cases the optimization is more important than runtime modulus flexibility...

One other option that's worth mentioning: we might look into how fast a modulus-flexible field library we could build, on the assumption that *for a given value* you wouldn't change the modulus very often. This would let us keep values in Montgomery form (almost) all the time, which gives a nice speed-up, but doesn't go all the way to requiring compile-time modulus selection. This option is compatible with the one directly above, in the sense that we could have this library implement the trait(s) that the IR requires and then plug this new library in (only?) when modulus flexibility is required.

kwantam:
This PR does three things:

- It adds a new type for field elements. This type can support multiple fields and multiple implementations for each field; right now, it supports `ff`-based fields for the BLS12-381 and BN254 scalar fields, plus a `rug`-based field for arbitrary moduli.
  - this type is defined in `circ_fields`. the reason is that deriving `ff` fields is very slow, so if it lived in the main crate then `cargo check` and friends would be extremely slow. as-is, `circ_fields` gets recompiled pretty infrequently, which is nicer for development.
  - there's a new feature, `ff_dfl`, that makes `ff`-based fields the default for CirC-IR. This is off by default. In my tests, the cost of converting back and forth to Integers (say, because of conversions to/from bitvectors) more than offset the savings from using `ff` instead of `rug` in the IR (as @alex-ozdemir suspected it would). But I also added code in the R1CS back-end that supports switching from Integer to ff-based field elements (this happens automatically in the `zxc` and `circ` examples), so the r1cs back-end uses ff.

- it pulls in Alex's changes in #60. This helps to avoid quadratic blow-ups that can happen in some cases.

- it also includes a few small fixes in the zsharp front-end 

closes #31


kwantam:
Hm, @edwjchen @alex-ozdemir has something broken with ABY recently? The tests appear to fail because the ABY interpreter is rejecting the `-p` commandline flag. But as far as I can tell, this is not a recent change: it seems to be coming from `scripts/aby_tests/utils.py`, which seemingly hasn't changed since #47.

edwjchen:
@kwantam Yes! There is now a new ABY interpreter from [PR 65](https://github.com/circify/circ/pull/65) that doesn't require the `-p` flag anymore. Once the PR is merged in, the ci pipeline should pass.

kwantam:
So it's expected that things will fail? (But then how does the main branch pass?)

In any case, sounds like there's a way forward once #65 lands.

kwantam:
@alex-ozdemir fyi there appears to be a memory leak in the linearity reduction code. Hunting it down. Let's hold on this PR until I've found it.

kwantam:
OK, it looks like it's small and related to Rust internals, so this is basically g2g.

It might still be useful to think about a more compact representation for linearity reduction, e.g., delay folding in linearity reductions. I might play with this tonight, but it's not clear to me how helpful this really is. I'll keep you posted.

src/ir/term/text/mod.rs
```diff
@@ -440,11 +451,9 @@ impl<'src> IrInterp<'src> {
...
                             "A set_default_modulus should have 2 arguments: modulus and term"
                         );
-                        let m = Arc::new(self.int(&tts[1]));
-                        self.moduli.push(m);
-                        let t = self.term(&tts[2]);
-                        self.moduli.pop().unwrap();
-                        t
+                        let m = self.int(&tts[1]);
+                        self.default_modulus = Some(m);
+                        self.term(&tts[2])
```
alex-ozdemir: I'm sorry Riad, I gave you bad information before. This was **not** a bug. The intended behavior is for `set_default_modulus` to be a *term* form that sets the default modulus within the scope of *that term*. So the pop is intentional.

I've pushed a commit that adds a note to the module documentation, reverts the change, and tests the right behavior.

On the other hand, if you think the new behavior is preferable, I'm happy to discuss.

Also, I like the Arc cache :)

alex-ozdemir (approved):
Generally, looks good. Let's re-run CI though, with the ABY unblocked; the ZoKrates integration tests fail on my machine.

alex-ozdemir:
> It might still be useful to think about a more compact representation for linearity reduction, e.g., delay folding in linearity reductions. I might play with this tonight, but it's not clear to me how helpful this really is. I'll keep you posted.
> 
> (Edit: to clarify, I mean that we might delay the weighted summations in `sub_in`. I'm hoping for a representation that's more space efficient without requiring too much more computation...)

I agree that this kind of change might help, but I'm also not sure.

src/ir/term/text/mod.rs
alex-ozdemir: Edit: While I thought I replied before saying this was a bug, apparently I didn't actually do that? Anyways, I'll just see myself out on my two left feet...

kwantam: Cool, no worries. I'll see what's up with the failures.

kwantam:
> Generally, looks good. Let's re-run CI though, with the ABY unblocked; the ZoKrates integration tests fail on my machine.

OK, I'll rebase and have a look.

Regarding SetDefaultModulus: I might add the Arc-cache back, but happy to leave the semantics as you intended.

kwantam:
> I agree that this kind of change might help, but I'm also not sure.

I might revisit later. The perf improvement just from your change is huge, and making it slightly faster gives only incremental returns on constraint count (at the cost of a lot more complexity).

kwantam:
Oh! You kept the arc cache. missed that at first. nice!

kwantam:
Rebased on master branch. Tests pass locally now, so I'm guessing things should be OK.

kwantam:
@alex-ozdemir from my perspective ready for merge. go for it if you agree!

PR #70, opened 2022-04-04
Spartan Backend

jkwoods:
Notes:
 - C -> r1cs -> spartan compilation is still halted by [this line](https://github.com/circify/circ/blob/8c8e7049146e1a326226ab8f04d6fd0b7a3d14f2/src/front/c/mod.rs#L658), which I'm assuming is in progress. If this line is commented out, works as expected for me.
 - run by using `--action spartan`
 - Let me know if you want me to make a circ field for the Spartan modulus/field. I didn't want to mess with defaults/don't know what the standard is for added things like this.
 - This is my first time writing rust, criticism welcome.


kwantam:
Thanks, this is awesome!

By the Spartan modulus, do you mean the field whose cardinality equals the Curve25519 group order?

@alex-ozdemir may have other thoughts, but I wonder if we shouldn't consider delaying this until we cut apart the codebase into separate modules (which I still think is the right thing to do!) and then make this one of those modules.

jkwoods:
Yes, that is exactly what I mean by Spartan modulus. Spartan uses the prime order group ristretto255 on top of Curve25519.

kwantam:
:+1: one or the other of us can add it in the `circ_fields` sub-crate

kwantam:
Hm, we probably also need to do a better job with field selection based on features.

I'm happy to look at the circ_fields edits. Sorry that I haven't fully macro-ified yet...

jkwoods:
Thanks, that would be great! I believe Curve25519 uses 9 as a generator.
I wasn't sure if you all wanted spartan (and bellman?) to be a separate feature?

edwjchen:
@jkwoods [this line](https://github.com/circify/circ/blob/8c8e7049146e1a326226ab8f04d6fd0b7a3d14f2/src/front/c/mod.rs#L658) was there because, at the time, I was using the C frontend to target the MPC backend. Feel free to remove the line if your tests pass!

alex-ozdemir (commented):
Looks good! In the long run, we might want to split the spartan action up, but for now, this is fine.

Re: the feature-gating, perhaps @rsw can do that when he adds the new field?

examples/circ.rs
```diff
@@ -317,6 +337,37 @@ fn main() {
...
+                ProofAction::Spartan => {
+                    println!("Converting R1CS to Spartan");
+ 
+                    let (inst, vars, inps, num_cons, num_vars, num_inputs) = r1cs_to_spartan(r1cs);
+
+                    println!("Proving with Spartan");
+                    let prover = Instant::now();
+
+                    // produce public parameters
+                    let gens = NIZKGens::new(num_cons, num_vars, num_inputs);
```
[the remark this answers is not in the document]
jkwoods: Yes, since we were not measuring a separate trusted setup. I'm sorry - I'll remove timing stuff.

Cargo.toml
```diff
@@ -36,6 +36,11 @@ pest-ast = "0.3"
 from-pest = "0.3"
 itertools = "0.10"
 petgraph = "0.6"
+spartan = "0.4.1"
+merlin = "3.0.0"
+curve25519-dalek = {version = "3.2.0", features = ["serde", "simd_backend"]}
+serde = { version = "1.0.124", features = ["derive"] }
+bincode = "1.3.3"
```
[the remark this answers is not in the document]
jkwoods: Yes. These are only added for Spartan (and therefore r1cs)

jkwoods:
The reason this isn't building I think is that Spartan by default turns on a `simd` feature (that only works with nightly) in their `curve25519-dalek` dependency. They say you can fix this but deleting it from their Cargo.toml, but I don't see how we can do that from the CirC level. (As disabling it in the CirC Cargo.toml isn't helpful - Spartan re-adds it.) I'm not sure what the elegant fix here is.

srinathsetty:
@jkwoods Please feel free to submit a PR to Spartan that puts `simd_backend` under a (default) feature. Circ can then include Spartan with default_features set to false.

jkwoods:
@srinathsetty Even with `simd_backend` under a default feature, I couldn't get Spartan to compile without using nightly. It uses feature create attributes (for features `test` and `int_log`), which I don't think can be removed/opted out of (unless I'm missing something about rust features).

jkwoods:
@alex-ozdemir @kwantam What if I gated all of the Spartan stuff behind a nightly only feature? Would that be an acceptable solution?

alex-ozdemir:
That would be fine with me. @kwantam ?

srinathsetty:
@jkwoods I've merged a few PRs that enable Spartan to run on stable Rust (without curve25519-dalek's simd_backend, so performance could be 50% slower for crypto ops). We plan to release a new crate soon. Please let me know if you run into any issues.

driver.py
```diff
@@ -14,7 +14,7 @@ def install(features):
         features : set of str
             set of features required
     """
-
+
```
alex-ozdemir, line 17:
```suggestion

```

alex-ozdemir (changes requested):
This is almost good to go. We'll need a few changes before merging:
* feature-gate appropriate stuff behind `r1cs`
* merge master back in
* remove ABY submodule

jkwoods:
Re: feature gating. The easiest thing to do, I think, would be to put the "spartan" features behind the "r1cs" feature (where "bellman" also is). This works well for something like c; in `examples/circ`, I just catch when we're in spartan mode [here](https://github.com/jkwoods/circ/blob/050eba384ef4d8cd5be4f2e0921eda9c5f2cfbff/examples/circ.rs#L292), and switch the field (spartan and bellman have different fields). This is sort of hacky though.

However, if you are using a front end like zokrates (and the spartan backend), you need the default field (DFL_T) to match. So it would make more sense to do away with the "r1cs" feature and split "spartan" and bellman" into separate features that each select a different default field. This would also be more elegant, but is a large overhaul. Would this be alright for me to do? Or overkill?

A third option is to put both "bellman" and "spartan" in "r1cs" and add an additional feature that you have to select for the spartan field. (And then if a user tries to run the spartan backend without selecting this feature, there's an error message.) This is maybe a good compromise?

alex-ozdemir:
> A third option is to put both "bellman" and "spartan" in "r1cs" and add an additional feature that you have to select for the spartan field. (And then if a user tries to run the spartan backend without selecting this feature, there's an error message.) This is maybe a good compromise?

Why don't we start with this? If all goes well, we can progress to the second approach later.

src/target/r1cs/spartan.rs
```diff
@@ -0,0 +1,161 @@
...
+pub fn r1cs_to_spartan<S: Eq + Hash + Clone + Display>(r1cs: R1cs<S>) -> (Instance, Assignment, Assignment, usize, usize, usize)
+{
+
+    // spartan format mapper: CirC -> Spartan
+    let mut wit = Vec::new();
+    let mut inp = Vec::new();
+    let mut trans: HashMap<usize, usize> = HashMap::new(); // Circ -> spartan ids
+    let mut itrans: HashMap<usize, usize> = HashMap::new(); // Circ -> spartan ids
+
+    match r1cs.values {
```
elefthei, line 38: Depracated API, see https://github.com/circify/circ/pull/80

jkwoods: fixed

jkwoods:
I think this is good to go now. Some notes:

1) I put all of the spartan stuff behind the `r1cs` feature. I added a `ristretto255` feature to select the spartan field. To make this work nicely in [src/util/field.rs](https://github.com/jkwoods/circ/blob/91f4677ec6cf2165f5b3b05a7d6429f2e651437c/src/util/field.rs#L1), I left the `bls12381` feature as the default and also added a `bn254` feature (see the [Cargo.toml](https://github.com/jkwoods/circ/blob/91f4677ec6cf2165f5b3b05a7d6429f2e651437c/Cargo.toml#L58)). This still seems over complicated to me, so if there's a better way to organize this let me know. (I also don't know exactly what the `ff_dfl` feature does - let me know if I handled that wrong.)

2) Speaking of fields, I tried to make the Spartan field into an ff circ_field, but was unable to, because ff_derive can't implement a square root function that works for this modulus I think (see [here](https://github.com/kwantam/ff/blob/b3041c177b9ae28095c9ce8347600212252abc83/ff_derive/src/lib.rs#L574)). So the Spartan field is just an IntField for now.

3) I made an example, `spartan_mul.zok`, with the Spartan modulus [here](https://github.com/jkwoods/circ/tree/dmz/examples/ZoKrates/pf). To run this example you would do:
```
cargo build --release --example circ --features "zok, smt, r1cs, ristretto255" --no-default-features
cargo build --release --example zk --features "r1cs, ristretto255" --no-default-features

./target/release/examples/circ examples/ZoKrates/pf/spartan_mul.zok r1cs --action spartansetup    # compile r1cs, write prover/verifier data
./target/release/examples/zk --pin examples/ZoKrates/pf/spartan_mul.zok.pin --vin examples/ZoKrates/pf/spartan_mul.zok.vin --action spartan     # read prover/verifier data, prove/verify

```
- Note that default features have to be turned off for correct compilation.

4) To make this example, and our workflows in general, run, I put a Spartan hook into examples/circ.rs [here](https://github.com/jkwoods/circ/blob/9ba2555bfa2beee2212bb83f3ac09dd948e58d37/examples/circ.rs#L315) and examples/zk.rs, [here](https://github.com/jkwoods/circ/blob/9ba2555bfa2beee2212bb83f3ac09dd948e58d37/examples/zk.rs#L54) (with associated command line flags), to mimic bellman. However, unlike Bellman, some of the structures that Spartan uses for proving/verifying (notably R1CS `Instances` and the `NIZKGens` proof parameters) can't be serialized and easily written to file. So I left the proving and verifying machinery in the same command. That being said, I have split up proving and verifying into separate functions in src/target/r1cs/spartan.rs ([here](https://github.com/jkwoods/circ/blob/91f4677ec6cf2165f5b3b05a7d6429f2e651437c/src/target/r1cs/spartan.rs#L21)), that each read their own input files, so it mirrors the same separation as bellman. This means I had to create two separate flags for prover and verifier inputs in examples/zk.rs.

Let me know if anything doesn't make sense and needs to be changed!

examples/circ.rs
```diff
@@ -28,7 +28,9 @@ use circ::target::aby::trans::to_aby;
 #[cfg(feature = "lp")]
 use circ::target::ilp::{assignment_to_values, trans::to_ilp};
 #[cfg(feature = "r1cs")]
-use circ::target::r1cs::bellman::{gen_params, prove, verify};
+use circ::target::r1cs::bellman::gen_params;
```
alex-ozdemir, line 31: Good catch! (removing prove/verify)

src/ir/term/mod.rs
```diff
@@ -1348,12 +1348,12 @@ impl Value {
         }
     }
     #[track_caller]
-    /// Get the underlying bit-vector constant, or panic!
+    /// Get the underlying integer constant, or panic!
```
alex-ozdemir, line 1351: Good catch.

examples/zk.rs
```diff
@@ -25,6 +30,7 @@ arg_enum! {
     enum ProofAction {
         Prove,
         Verify,
+        Spartan,
```
alex-ozdemir, line 33: If the `arg_enum` macro allows it, I'd add a doc-comment saying that this action does both proving and verifying.

src/util/field.rs
```diff
@@ -3,26 +3,42 @@
 // NOTE: when we eventually break CirC into separate crates,
 //       each crate may want its own field type default
```
alex-ozdemir, line 5: @kwantam Do you have time to look at the changes to this file? I think you have the best perspective.

driver.py
```diff
@@ -16,7 +16,7 @@ def install(features):
...
```
alex-ozdemir, line 19: Is this change intended?

Cargo.toml
```diff
@@ -54,10 +57,12 @@ approx = "0.5.0"
 [features]
 default = ["bls12381", "ff_dfl"]
 bls12381 = []
+bn254 = []
+ristretto255 = []
 ff_dfl = []
 c = ["lang-c"]
 lp = ["good_lp", "lp-solvers"]
-r1cs = ["bellman"]
+r1cs = ["bellman", "spartan", "merlin", "curve25519-dalek"]
```
alex-ozdemir, line 65: Great! I see that you followed our plan here.

---

Beyond this PR, we probably should do some feature clean-up soon. For features that we intend to be externally set (either when using CirC as a library, or when building one of its binaries), I'd propose the following principle: **each feature should correspond to *one* backend or frontend**.

Thus, we'll want to make `r1cs` an internal feature (implied by other features, and not exposed in `driver.py`). It will be implied by either of `bellman` and `spartan` (the latter will imply `merlin` and `curve25519-dalek`).

We don't need to make that change in this PR, but I bring it up b/c I'd like to know your opinion of this idea, Jess. Also, @kwantam.

alex-ozdemir (changes requested):
First, *thank you*, *thank you*, *thank you* Jess! This is great! :tada: 

Before merging:
* you should address a few tiny comments I left
* We need @kwantam to review the default field changes.

Then we're good to go in my book. I also left a question in-line for you. I very much care about your answer and @kwantam 's answer, but we won't hold up the merge on it.

A final question: does the `dmz` branch name mean "de-militarized zone" or something else?

driver.py
jkwoods: no, oops!

jkwoods:
No problem! dmz does indeed mean "de-militarized zone".

Cargo.toml
jkwoods: I like this idea. What about in the case when you might want to compile R1CS without sending it immediately to a "built-in" proving system (like count)? I guess you would need to specify a modulus anyway (in most cases). So it would be okay to have `r1cs` as well as features implying different fields (`bls12381`, `ristretto255`, etc) implied by `bellman`/`spartan` features. Then you could specify, for example, `bellman` even if you weren't proving, to say essentially "use the bellman modulus to create an R1CS instance". I'm not sure if this is more confusing though.

examples/zk.rs
jkwoods: Added. `arg_enum` allows one comment for the enum, but not individual ones for different options.

alex-ozdemir (approved):
Looks good to me! We'll see if rsw says anything.

elefthei:
Can we merge this? If there are changes we can always do another PR.

alex-ozdemir:
Yeah, I think @kwantam  is probably too busy right now. If CI goes through, I'm happy.

Thanks for bumping this.

jkwoods:
Oh I'm sorry! I fixed a bug having to do with splitting up prover and verifier input files and maintaining order for Spartan, and now can't see all the formatting stuff I need to fix until CI is rerun.

jkwoods:
In addition to ci formatting stuff, I fixed a bug arising from splitting up prover and verifier inputs and needing to maintain the same input/output order for both.

alex-ozdemir:
We're making some progress. It seems like there are a few lint issues. (`./driver -l`)

jkwoods:
Sorry, I haven't used real CI like this before, had to figure it out.

jkwoods:
I saw these lint issues when I ran linting locally (before _and_ after last commit), but they didn't show up on the CI before and they aren't in code I touched, so I thought it wouldn't matter. It doesn't seem to be related to my changes. Any idea what's going on?

jkwoods:
No problem! Some questions:
1. `ctx` doesn't seem to be used in [this function](https://github.com/circify/circ/blob/90f6b2ff354a0e170c9bff444b8edb1147d96618/src/circify/mod.rs#L928), which clippy is not fond of. Can I get rid of it, or should I leave it as `_ctx` (for debugging, etc)?
```
            fn ite(&self, ctx: &mut CirCtx, cond: Term, t: Self::T, f: Self::T) -> Self::T {
                match (t, f) {
                    (T::Base(a), T::Base(b)) => T::Base(term![Op::Ite; cond, a, b]),
                    (T::Pair(a0, a1), T::Pair(b0, b1)) => T::Pair(
                        Box::new(self.ite(ctx, cond.clone(), *a0, *b0)),
                        Box::new(self.ite(ctx, cond, *a1, *b1)),
                    ),
                    (a, b) => panic!("Cannot ITE {}, {}", a, b),
                }
            }
```

2. Clippy tags some functions like [this](https://github.com/circify/circ/blob/90f6b2ff354a0e170c9bff444b8edb1147d96618/src/front/c/mod.rs#L470) because `self` is "only used in recursion". I see why this could be an issue for other parameters from looking at their [docs](https://rust-lang.github.io/rust-clippy/master/index.html#expect_fun_call) but it doesn't seem feasible to get rid of `self`. No one else seems to have run into this problem. Can I just tell clippy to ignore this?
```
fn base_loc(&self, loc: CLoc) -> CLoc {
        match loc {
            CLoc::Var(_) => loc,
            CLoc::Member(l, _) => self.base_loc(*l),
            CLoc::Idx(l, _) => self.base_loc(*l),
        }
    }
```
I'm sorry for the extra hassle.

alex-ozdemir:
No problem! Some questions:

> ctx doesn't seem to be used ... Can I get rid of it, or should I leave it as
> _ctx (for debugging, etc)?

Yep, `_ctx` is good here. `ite` is a trait function, so you can't change the
signature (other implementation may need `ctx`).

> `base_loc`

Yeah, this seems like a false positive. I'd allow it, and leave a comment.

jkwoods:
Sorry to just think of this, but should I add workflow tests for spartan in `scripts/`, or is that unnecessary/can wait?

alex-ozdemir:
> Sorry to just think of this, but should I add workflow tests for spartan in scripts/, or is that unnecessary/can wait?

It's your call. If you want the spartan back-end to continue to work long-term, I'd add a test for it :). It's easy to break stuff accidentally without tests.

jkwoods:
Okay - wrote some test infastructure that mirrors the zok -> bellman testing pretty closely. Because testing spartan requires the recompilation of CirC with a different DFL_T, I had to add in some stuff to `driver.py` and `.github/workflows/ci.yml`. I did this in what I think was the most straightforward way, but let me know if I'm messing up y'all's flow.

.github/workflows/ci.yml
```diff
@@ -41,3 +41,13 @@ jobs:
...
+    - name: Set spartan field feature on
+      run: python3 driver.py --features zok smt r1cs ristretto255
+    - name: Check (spartan custom build)
+      run: python3 driver.py --check
+    - name: Format (spartan custom build)
+      run: cargo fmt -- --check
+    - name: Lint (spartan custom build)
+      run: python3 driver.py --lint
+    - name: Build, then Test (spartan custom build)
+      run: python3 driver.py --test
```
alex-ozdemir, line 53: Possibly naive question: is it possible for us to run both the bellman and spartan tests from a single build (with a single set of enabled features)?

jkwoods: Not with the way the fields are currently set up. The choice of field is made at build time (selected in `src/util/field.rs` by chosen features). This field can technically be disregarded/changed after building, right before conversion of r1cs to spartan, if you are compiling a front end like C. But for a front end like Zokrates, that encodes literal field elements, it is important the field is correct from the get go.

So you can either select the bellman field OR the spartan field during the build.

alex-ozdemir: Okay, that makes sense. I think we could get around this restriction, but it'll take a bit of work.

I worry about the latency & potential cache-thrash induced by doing a double-build. I propose that this PR should merge the test script and driver modifications, but not modify the CI script. When we fix the field setup, we can broaden CI.

Does sound good to you?

jkwoods: Sure!

alex-ozdemir:
Yes!

PR #73, opened 2022-04-07
zsharp fixes ; speed improvements

kwantam:
This PR does two things:

- Fixes some scoping issues in Z# having to do with the context in which function and struct definitions are interpreted.
- Improves speed and reduces TermTable memory consumption by 2x (by keying on Arc<TermData> rather than TermData)
- Improves speed for Z# array handling with a BTreeMap -> HashMap swap
- Adds intrinsic widening casts in Z#. These were previously handled by bit-splitting, which is wasteful.
 
On array-heavy Z# programs, this PR improves speed by about 5x and memory consumption by roughly the same.

(EDIT: mentioned one other thing this PR does)

### DIFF
--- a/src/front/zsharp/term.rs
+++ b/src/front/zsharp/term.rs
@@ -772,7 +772,7 @@ pub fn array_store(array: T, idx: T, val: T) -> Result<T, String> {
 }
 
 fn ir_array<I: IntoIterator<Item = Term>>(sort: Sort, elems: I) -> Term {
-    let mut values = BTreeMap::new();
+    let mut values = HashMap::new();
     let to_insert = elems
         .into_iter()
         .enumerate()
@@ -791,7 +791,7 @@ fn ir_array<I: IntoIterator<Item = Term>>(sort: Sort, elems: I) -> Term {
     let arr = leaf_term(Op::Const(Value::Array(Array::new(
         Sort::Field(DFL_T.clone()),
         Box::new(sort.default_value()),
-        values,
+        values.into_iter().collect::<BTreeMap<_, _>>(),
         len,
     ))));
     to_insert
--- a/src/front/zsharp/term.rs
+++ b/src/front/zsharp/term.rs
@@ -2,7 +2,6 @@
 use std::collections::{BTreeMap, HashMap};
 use std::fmt::{self, Display, Formatter};
 
-use log::warn;
 use rug::Integer;
 
 use crate::circify::{CirCtx, Embeddable};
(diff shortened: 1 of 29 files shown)


src/ir/term/mod.rs
```diff
@@ -944,7 +944,7 @@ pub type Term = HConsed<TermData>;
 pub type TTerm = WHConsed<TermData>;
 
 struct TermTable {
-    map: FxHashMap<TermData, TTerm>,
+    map: FxHashMap<Arc<TermData>, TTerm>,
```
alex-ozdemir, line 947: Excellent change.

src/front/zsharp/term.rs
```diff
@@ -772,7 +769,7 @@ pub fn array_store(array: T, idx: T, val: T) -> Result<T, String> {
 }
 
 fn ir_array<I: IntoIterator<Item = Term>>(sort: Sort, elems: I) -> Term {
-    let mut values = BTreeMap::new();
+    let mut values = HashMap::new();
```
alex-ozdemir, line 772: So I assume the speedup here is in the case that there are a lot of constants in the array?

src/front/zsharp/zvisit/zgenericinf.rs
```diff
@@ -352,6 +359,7 @@ impl<'ast, 'gen, const IS_CNST: bool> ZGenericInf<'ast, 'gen, IS_CNST> {
         // 2. walk through struct def to generate constraints on inner explicit generics
         let old_sfx = std::mem::replace(&mut self.sfx, new_sfx);
         let old_gens = std::mem::replace(&mut self.gens, &strdef.generics[..]);
+        self.zgen.file_stack_push(strpath);
         for ast::StructField { ty, id, .. } in strdef.fields.iter() {
             if let Some(t) = aty_map.remove(&id.value) {
                 self.fdef_gen_ty(t, ty)?;
```
alex-ozdemir, line 365: Oh interesting. I guess we don't fully resolve types at declaration time?

alex-ozdemir (approved):
This looks good. I left two questions, just for my own understanding

src/front/zsharp/term.rs
kwantam: Yeah, exactly. Large constant arrays were a big speed issue previously.

src/front/zsharp/zvisit/zgenericinf.rs
kwantam: Right. Since we have to do generic inference at "call time" anyhow, I just leave all the resolution until then. I guess we could resolve "real" constants at declaration and delay resolving generics until inference time, but as far as I can tell that just adds complexity.

PR #74, opened 2022-04-10
Improve non-recursive type-checking

alex-ozdemir:
In *non-recursive* type-checking, perhaps better called *type computing*
we perform a minimal traversal in order to compute the type of a term,
without recursively type-checking it. Informally, we assume it is well
typed, and do the minimal amount of work needed to compute its type.

Two improvements:
1. No longer implemented with recursion.
2. Caches all intermediate results.

Implementation:
* `check_dependencies(Term) -> Vec<Term>`: maps a terms to the immediate
  children of it whose types are needed to compute its type.
* `check_raw_step(Term, TypeTable) -> Sort`: assumes those children have
  their types in the table, and computes this term's type.
* `check_raw`: glues the two above functions together into a suitable
  traversal. Similar to `rec_check_raw`, but the traversal isn't total.

Significance:

Previously, we had a non-recursive implementation for array stores that
*didn't cache intermediate results* which could cause quadratic
type-checking time. Edward sent me a benchmark that was experiencing 
this, resulting in 74.1% of total compilation time being spent type-checking.
Now it's down to 0.4% of total compilation time.

Before:
flamegraph

After:
flamegraph

### DIFF
--- a/examples/C/mpc/benchmarks/2pc_biomatch_40x.c
+++ b/examples/C/mpc/benchmarks/2pc_biomatch_40x.c
@@ -1,0 +1,38 @@
+int main(
+    __attribute__((private(0))) int db[40960],
+    __attribute__((private(1))) int sample[4])
+{
+    int N = 10240;
+    int K = 4;
+    int matches[N];
+
+    // Compute distances
+    for (int i = 0; i < N; i++)
+    {
+        int db_inner[K];
+        for (int j = 0; j < K; j++)
+        {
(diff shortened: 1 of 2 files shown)


alex-ozdemir:
@edwjchen 

I started poking at your benchmark, and noticed that there was a huge *runtime* bottleneck that this PR fixes. I'll take a cut at the memory bottleneck next.

edwjchen (approved):

PR #77, opened 2022-04-25
improve memory consumption ; add Z# strict unary op, ternary ?:, and type aliases

kwantam:
This PR breaks cycles between the term cache and the constant folding machinery. It also adds a couple Drop implementations that explicitly call the garbage collector (the C front-end might want one, too---not sure!)

For Z#, this adds

- a unary operator `#` that strictly evaluates (i.e., constant folds) its argument. As a special case, an assignment's RHS is treated strictly if its LHS is `#( ... )`
- the `condition ? consequent : alternative` flavor of ternary operator - exactly the same as `if foo then bar else baz fi`
- typedefs: `type Foo = field[7]` or whatever.

Typedefs have one limitation: if a typedef `Bar` refers to a struct `Foo` in module A, then to use `Bar` in module B one must also import `Foo`.

@alex-ozdemir: before we land this PR, we should land https://github.com/alex-ozdemir/hashconsing/pull/2 and then I should update Cargo.lock to point to your repo again.

### DIFF
--- a/src/ir/term/mod.rs
+++ b/src/ir/term/mod.rs
@@ -999,7 +999,7 @@ impl TermTable {
             let okv = self.map.get_key_value(&*t.elm);
             std::mem::drop(t);
             if let Some((key, _)) = okv {
-                if Arc::strong_count(key) == 1 {
+                if Arc::strong_count(key) <= 1 {
                     to_check.extend(key.cs.iter().cloned());
                     let key = key.clone();
                     self.map.remove(&key);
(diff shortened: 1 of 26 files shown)


src/ir/term/mod.rs
```diff
@@ -1062,8 +1064,16 @@ fn mk(elm: TermData) -> Term {
 
 /// Scans the term database and the type database and removes dead terms and types.
 pub fn garbage_collect() {
-    collect_terms();
-    collect_types();
+    // Don't garbage collect while panicking.
+    // this function may be called from Drop implementations, which are called
+    // when a thread is unwinding due to a panic. When that happens, RwLocks are
+    // poisoned, which would cause a panic-in-panic, no bueno.
```
alex-ozdemir, line 1070: Nice!

alex-ozdemir (approved):
Looks good to me, pending the manifest reversion.

PR #78, opened 2022-04-26
Exposing vector_add to zok frontend

edwjchen:


src/front/zsharp/mod.rs
```diff
@@ -305,6 +305,25 @@ impl<'ast> ZGen<'ast> {
...
+                } else if generics.len() != 1 {
+                    Err(format!(
+                        "Got {} generic args to EMBED/vector_add, expected 1",
+                        generics.len()
+                    ))
+                } else {
+                    assert!(args.iter().all(|t| {
+                        let t_sort = t.type_();
+                        matches!(t_sort, Ty::Array(_, _))
+                    }));
```
kwantam, line 323:
```suggestion
                    assert!(args.iter().all(|t| matches!(t.type_(), Ty::Array(_, _))));
```

src/front/zsharp/mod.rs
```diff
@@ -305,6 +305,25 @@ impl<'ast> ZGen<'ast> {
...
+                    vector_op(BV_ADD, args[0].clone(), args[1].clone())
```
kwantam, line 324: These clones are wasteful. Just deconstruct the vector.

kwantam (commented):
A few comments so far. I'll have another look once you've got edits to the constant folder!

third_party/ZoKrates/zokrates_stdlib/stdlib/EMBED.zok
```diff
@@ -75,3 +75,7 @@ def u16_to_u32(u16 i) -> u32:
 
 def u8_to_u16(u8 i) -> u16:
     return 0u16
+
+// vector functions
+def vector_add<N>(u32[N] a, u32[N] b) -> u32[N]:
+    return [0; N]
```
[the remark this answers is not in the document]
edwjchen: I think only BV are currently supported in the FHE backend, which is why I left it as u32.

src/ir/opt/cfold.rs
```diff
@@ -244,6 +245,35 @@ pub fn fold_cache(node: &Term, cache: &mut TermCache<Term>) -> Term {
                     b.width() + w,
                 ))))
             }),
+            Op::Map(op) => match (get(0).as_array_opt(), get(1).as_array_opt()) {
+                (Some(a), Some(b)) => {
+                    assert!(a.size == b.size);
+                    let mut new_map: BTreeMap<Value, Value> = BTreeMap::new();
+                    for (a_ind, a_val) in &a.map {
+                        let b_val = &b.map[a_ind];
+                        let res = fold_cache(
```
edwjchen, line 254: @kwantam How would I evaluate the operation on `va` and `vb` without the recursive call to fold_cache here?

third_party/ZoKrates/zokrates_stdlib/stdlib/EMBED.zok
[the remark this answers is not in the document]
kwantam: Still, it might make sense to support all the basic types, no? It wouldn't be much extra work:

- add vadd_field, vadd_u64, vadd_u32, vadd_u16, vadd_u8 stubs in EMBED
- in builtin_call_, extend the current match clause to catch all of these
- extend the function in `zsharp/term.rs` to support both bitvector and field types
- the current constant folder may or may not work, but I bet there's a small edit distance to working if no

Since this is so simple, I don't see a reason to leave it half-done. But I could be missing something!

Cargo.lock
```diff
@@ -138,7 +138,7 @@ dependencies = [
  "log",
  "num_cpus",
  "pairing",
- "rand_core",
+ "rand_core 0.6.3",
```
kwantam, line 141: Why would Cargo.lock have changed? There are no new crates being included.

src/front/zsharp/mod.rs
```diff
@@ -305,6 +305,24 @@ impl<'ast> ZGen<'ast> {
                     Ok(uint_lit(DFL_T.modulus().significant_bits(), 32))
                 }
             }
+            "vector_add" => {
```
kwantam, line 308: I think this name should specify the type it handles, because it's not generic.

Maybe vadd_u32 or vec_add_u32 or addvec_u32?

src/front/zsharp/mod.rs
```diff
@@ -305,6 +305,24 @@ impl<'ast> ZGen<'ast> {
...
+                    ))
+                } else if generics.len() != 1 {
+                    Err(format!(
+                        "Got {} generic args to EMBED/vector_add, expected 1",
+                        generics.len()
+                    ))
+                } else {
+                    assert!(args.iter().all(|t| matches!(t.type_(), Ty::Array(_, _))));
+                    let a = args.pop().unwrap();
+                    let b = args.pop().unwrap();
```
kwantam, line 322:
```suggestion
                    let b = args.pop().unwrap();
                    let a = args.pop().unwrap();
```

`pop` takes from the end of the array, so you're reversing the order of arguments here.

src/front/zsharp/term.rs
```diff
@@ -908,6 +908,20 @@ pub fn bit_array_le(a: T, b: T, n: usize) -> Result<T, String> {
     ))
 }
 
+pub fn vector_op(op: Op, a: T, b: T) -> Result<T, String> {
+    match (&a.ty, &b.ty) {
+        (Ty::Array(a_s, a_ty), Ty::Array(b_s, b_ty)) => {
+            if a_s == b_s && a_ty == b_ty {
+                let t = term![Op::Map(Box::new(op)); a.term, b.term];
+                Ok(T::new(Ty::Array(*a_s, a_ty.clone()), t))
```
kwantam, line 916:
```suggestion
    match (a.ty, b.ty) {
        (Ty::Array(a_s, a_ty), Ty::Array(b_s, b_ty)) => {
            if a_s == b_s && a_ty == b_ty {
                let t = term![Op::Map(Box::new(op)); a.term, b.term];
                Ok(T::new(Ty::Array(a_s, a_ty), t))
```

There's no need to clone here. `a` and `b` are owned by this function.

src/ir/opt/cfold.rs
kwantam: High-level answer: the `Op::Map(op)` match clause starting on line 248 should push a new term onto the stack that is equivalent to the term containing the `Op::Map(_)`, but which eliminates the map. I can give you more details later.

src/ir/opt/cfold.rs
```diff
@@ -244,6 +245,35 @@ pub fn fold_cache(node: &Term, cache: &mut TermCache<Term>) -> Term {
...
+                    let new_arr = Array::new(
+                        a.key_sort.clone(),
+                        Box::new(a.key_sort.default_value()),
+                        new_map,
+                        a.size,
+                    );
+                    Some(leaf_term(Op::Const(Value::Array(new_arr))))
+                }
+                _ => None,
+            },
```
kwantam, line 276: As we discussed yesterday, this code doesn't work yet, right?

kwantam (commented):
a few more comments

third_party/ZoKrates/zokrates_stdlib/stdlib/EMBED.zok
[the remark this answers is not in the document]
edwjchen: I see, updated!

kwantam:
Sorry for the delay. I'll spend some time reviewing this today, I hope.

src/ir/opt/cfold.rs
```diff
@@ -244,6 +244,29 @@ pub fn fold_cache(node: &Term, cache: &mut TermCache<Term>) -> Term {
                     b.width() + w,
                 ))))
             }),
+            Op::Map(op) => match (get(0).as_array_opt(), get(1).as_array_opt()) {
+                (Some(a), Some(b)) => {
+                    // TODO: extend for n-ary arrays
+                    let mut res = a.clone();
+                    let mut merge = ArrayMerge::new(a.clone(), b.clone());
```
kwantam, line 251: Cloning is unnecessary here. ArrayMerge is just reading values (and the default value) out of `a` and `b`, which means it should be able to take a shared reference to each.

src/ir/opt/cfold.rs
```diff
@@ -244,6 +244,29 @@ pub fn fold_cache(node: &Term, cache: &mut TermCache<Term>) -> Term {
...
+                    for (i, va, vb) in merge.into_iter() {
+                        let r = fold_cache(
+                            &term![*op.clone(); leaf_term(Op::Const(va.clone())), leaf_term(Op::Const(vb.clone()))],
+                            cache,
+                        );
+                        match r.as_value_opt() {
+                            Some(v) => {
+                                res = res.clone().store(i, v.clone());
```
kwantam, line 259: This line clones `res` on every pass through the loop. That is quadratic time for totally reasonable values of `a` and `b`. And then the previous copy just gets thrown away!

```suggestion
                                res = res.store(i, v.clone());
```

src/ir/opt/cfold.rs
```diff
@@ -244,6 +244,29 @@ pub fn fold_cache(node: &Term, cache: &mut TermCache<Term>) -> Term {
...
                 ))))
             }),
+            Op::Map(op) => match (get(0).as_array_opt(), get(1).as_array_opt()) {
+                (Some(a), Some(b)) => {
+                    // TODO: extend for n-ary arrays
+                    let mut res = a.clone();
+                    let mut merge = ArrayMerge::new(a.clone(), b.clone());
+                    for (i, va, vb) in merge.into_iter() {
+                        let r = fold_cache(
+                            &term![*op.clone(); leaf_term(Op::Const(va.clone())), leaf_term(Op::Const(vb.clone()))],
```
kwantam, line 254: As currently written 3 lines above, ArrayMerge owns the `a` and `b` data. So it seems like ArrayMerge could return `va` and `vb` by value (because it owns those values and doesn't need them any more)---yet they're cloned again here. That seems wasteful.

I'd suggest having ArrayMerge take `a` and `b` by shared reference, per my prior comment, and either

1. have it return `va` and `vb` as references, in which case they'd be cloned here, or
2. have ArrayMerge clone `va` and `vb` and return them as values, in which case no need to clone here.

All else equal, it would be best if ArrayMerge returned references, because that's the most generic thing: if the caller needs values (as here), it can clone; but if it only needs references there hasn't been any time or memory wasted on cloning.

src/ir/opt/cfold.rs
```diff
@@ -244,6 +244,29 @@ pub fn fold_cache(node: &Term, cache: &mut TermCache<Term>) -> Term {
                     b.width() + w,
                 ))))
             }),
+            Op::Map(op) => match (get(0).as_array_opt(), get(1).as_array_opt()) {
+                (Some(a), Some(b)) => {
+                    // TODO: extend for n-ary arrays
+                    let mut res = a.clone();
```
kwantam, line 250: Why clone `a` here? Just create a new array. There's no need to copy all of the data inside of `a` only to throw it away a few lines later. Why not something like:

```suggestion
                    let mut res = Array::new(a.key_sort.clone(), a.default.clone(), Default::default(), a.size);
```

src/ir/term/mod.rs
```diff
@@ -721,6 +723,104 @@ impl Array {
         self.check_idx(idx);
         self.map.get(idx).unwrap_or(&*self.default).clone()
     }
+
+    /// Iter
+    pub fn into_iter(&self) -> std::collections::btree_map::IntoIter<Value, Value> {
```
kwantam, line 728: If you're going to implement the `into_iter` function, you should implement the `std::iter::IntoIterator` trait, not just add a member function.

Continuing my crusade against unnecessary cloning: you probably want `impl<'a> IntoIterator for &'a Array`, i.e., *not* a consuming iterator---just one that returns references. So I think what you really want is something like this (untested---see `impl<'a, K, V> IntoIterator for &'a BTreeMap<K, V>` in [the stdlib](https://doc.rust-lang.org/src/alloc/collections/btree/map.rs.html#1322)):

```rust
impl<'a> IntoIterator for &'a Array {
    type Item = (&'a Value, &'a Value);
    type IntoIter = std::collection::btree_map::Iter<'a, Value, Value>;
    fn into_iter(self) -> Self::IntoIter {
        self.map.iter()
    }
}
```

kwantam: A second, higher-level comment: it's extremely un-idiomatic for an `into_iter` function to take `&self` rather than `self`. This may seem slightly counterintuitive given what I've just said in my prior comment, but notice in my proposed `impl IntoIterator` that the argument to `fn into_iter` *is* `self`---it's just that we think of `self` as being a *value* of type `&'a Array`.

In contrast, what is currently implemented in the PR takes `&self` and then clones it to produce a by-value iterator, which is bad because it means that if someone did

```rust
    for (k, v) in some_array.into_iter() {
    }
```

they will have potentially caused a very expensive clone that they didn't expect (because idiomatically, into_iter never clones---it either consumes a value or it takes a reference and returns references).

Does this make sense?

src/ir/term/mod.rs
```diff
@@ -721,6 +723,104 @@ impl Array {
...
+        self.map.clone().into_iter()
+    }
+}
+
+/// Merge two Array Iterators
+pub struct ArrayMerge {
```
kwantam, line 734: Repeating from above: ArrayMerge should take references to arrays, not values. Even `left_dfl` and `right_dfl` should probably be `&Value`, not `Value`.

src/ir/term/mod.rs
```diff
@@ -721,6 +723,104 @@ impl Array {
...
+                a.key_sort, b.key_sort
+            );
+        }
+        if a.default.sort() != b.default.sort() {
+            panic!(
+                "IR Arrays default values have different key sorts: {}, {}",
+                a.default.sort(),
+                b.default.sort()
+            );
+        }
```
kwantam, line 759: `assert!()` is more compact than if-`panic!` and means exactly the same thing. You can even [add error messages](https://doc.rust-lang.org/std/macro.assert.html#custom-messages) in the same way.

src/ir/term/mod.rs
```diff
@@ -721,6 +723,104 @@ impl Array {
...
+
+        Self {
+            left: Box::new(a.into_iter()).peekable(),
+            right: Box::new(b.into_iter()).peekable(),
```
kwantam, line 763: I doubt you need `Box` here even in the current implementation. You would need a `Box` if you wanted to make this generic over any iterator (because in that case it would be a trait object, `Box<dyn ...>` or so). But here you're referring to a concrete type that's sized, so you definitely don't need to box it, and you don't need the extra level of indirection that comes with a heap-allocated object.

This will be true after converting to work with references and implementing `IntoIterator` as described above---Array.into_iter() returns a concrete type (`std::collection::btree_map::Iter<'a, Value, Value>`), and that's the type that this struct will hold.

src/ir/term/mod.rs
```diff
@@ -721,6 +723,104 @@ impl Array {
...
+            left_dfl: *a.default,
+            right_dfl: *b.default,
+        }
+    }
+
+    /// Iter
```
kwantam, line 769: This is not right.

1. An iterator should be *lazy*. Here, you're running the whole iterator at once, putting the result into a vector, and then turning that vector into an iterator.
2. As above, `into_iter` is not something you implement as a member function. If you need `into_iter`, you implement the `IntoIterator` trait. But *you do not need to do that here*. More below.

The `ArrayMerge` struct should implement `std::iter::Iterator` directly. There's no need to implement `IntoIterator`. The `next_` function you've written should be part of the `impl Iterator` for `ArrayMerge`, not a member function. And, per the above, it should return something like `Option<(&'a Value, &'a Value, &'a Value)>`.

The functionality inside `next_` seems mostly right, it's just the way that you're connecting the machinery together that isn't right. So fixing this should be pretty easy!

When in doubt, read stdlib code.

src/ir/term/mod.rs
```diff
@@ -721,6 +723,104 @@ impl Array {
...
+
+    /// Next
+    pub fn next_(&mut self) -> Option<(Value, Value, Value)> {
+        let l_peek = self.left.peek();
+        let r_peek = self.right.peek();
+
+        let mut left_next = false;
+        let mut right_next = false;
+
+        let res = match (l_peek, r_peek) {
```
kwantam, line 788: Wouldn't it be simpler to say

```suggestion
        let res = match (self.left.peek(), self.r.peek()) {
```

and get rid of lines 782-783? They don't help readability.

src/ir/term/mod.rs
```diff
@@ -721,6 +723,104 @@ impl Array {
...
+        Box::new(acc.into_iter())
+    }
+
+    /// Next
+    pub fn next_(&mut self) -> Option<(Value, Value, Value)> {
+        let l_peek = self.left.peek();
+        let r_peek = self.right.peek();
+
+        let mut left_next = false;
+        let mut right_next = false;
```
kwantam, line 786: These are unnecessary. See below.

src/ir/term/mod.rs
```diff
@@ -721,6 +723,104 @@ impl Array {
...
+
+        let res = match (l_peek, r_peek) {
+            (Some((l_ind, l_val)), Some((r_ind, r_val))) => match l_ind.cmp(r_ind) {
+                Ordering::Less => {
+                    left_next = true;
+                    Some((l_ind.clone(), l_val.clone(), self.right_dfl.clone()))
```
kwantam, line 792: More unnecessary cloning here, only to throw away `self.l.next()` below.

The right way to do this is to call `self.l.next()` right here and *use that value*. That also lets you get rid of the mutable state variables `left_next` and `right_next`, as I mentioned above.

```suggestion
                    let (l_ind, l_val) = self.l.next().unwrap(); // unwrap is safe because we peeked
                    Some((l_ind, l_val, self.right_dfl.clone()))
```

Of course, after fixing things `self.right_dfl` won't be cloned, it'll just be a reference.

kwantam: Of course, same comments apply to the remaining match clauses.

kwantam (changes requested):
I left a few comments. Happy to jump on a call and discuss in more detail. The upshot is, the functionality looks mostly right but what's currently implemented is unnecessarily inefficient.

PR #81, opened 2022-05-07
concurrency fixes for garbage collection

kwantam:
@alex-ozdemir reminded me that tests run concurrently, meaning we actually have to worry about deadlocking etc in the garbage collector. Mea culpa.

This PR makes the collector hold one lock (on TERMS, TYPES, or FOLDS) at a time.

It also adds a new lock, COLLECT, only in `#[cfg(test)]`. The collector takes this read-write; constant-folding and type-checking take it read-only. This means that garbage collection can't run while folding or typing is going on, which otherwise could potentially result in terms being collected while they're still needed.


alex-ozdemir (approved):
Looks good, modulo the lock tweaks.

PR #106, opened 2022-09-19
Clone audit of the C frontend

edwjchen:
- Pass function parameters by reference when applicable. 
- Removing unnecessary clones.


alex-ozdemir:
Can you say more about the purpose of this PR?

From the title, I thought you might just be stripping a few un-needed clones,
but the patch seems much bigger than that?

alex-ozdemir (approved):

PR #107, opened 2022-10-01
Rotation IR intrinsic for SIMD FHE operations

edwjchen:


alex-ozdemir:
A few questions to start with:
* are we interested in symbolic rotations, or is the rotation amount always concrete?
* is there a reason that this PR doesn't define evaluation for `Op::Morph`?

edwjchen:
> * are we interested in symbolic rotations, or is the rotation amount always concrete?
I believe for the FHE backend, rotation amounts are always by a public constant. As far as I can tell, I'm don't think we need support for symbolic rotations (yet?). 

> * is there a reason that this PR doesn't define evaluation for `Op::Morph`?
Morph would a public permutation function on a packed ciphertext, but the permutation depends on the number of elements in the ciphertext and the ciphertext's modulus. I dont think we have a clear way of encoding this information yet, we could look to table the Morph intrinsic for now?

alex-ozdemir:
> I believe for the FHE backend, rotation amounts are always by a public
> constant. As far as I can tell, I'm don't think we need support for symbolic
> rotations (yet?).

Let's figure this out. If we don't need symbolic rotations, it'd be better to
make the rotation operator a one-argument operator indexed by a constant (like
our [bit-vector extension operators][1])

> Morph would a public permutation function on a packed ciphertext, but the
> permutation depends on the number of elements in the ciphertext and the
> ciphertext's modulus. I dont think we have a clear way of encoding this
> information yet, we could look to table the Morph intrinsic for now?

Yes, in general, I think that we shouldn't include any operator we don't know
how to evaluate yet.

Another question about rotation: is rotation well defined on an array whose
size is **not** equal to the size of the key space? For example, on an array
keyed by 4-bit bit-vectors, with a size of only 15?

[1]: https://github.com/circify/circ/blob/master/src/ir/term/mod.rs#L76-L79

edwjchen:
> Let's figure this out. If we don't need symbolic rotations, it'd be better to
make the rotation operator a one-argument operator indexed by a constant (like
our [bit-vector extension operators](https://github.com/circify/circ/blob/master/src/ir/term/mod.rs#L76-L79))

I just confirmed that rotation indices are public, so I'll make the change to match that of bit-vector extension operators. 

> Another question about rotation: is rotation well defined on an array whose
size is not equal to the size of the key space? For example, on an array
keyed by 4-bit bit-vectors, with a size of only 15?

In existing FHE compilers, arrays are all sizes with power-of-2. If an array is initialized with less values than a power-of-2, then the elements in the array are replicated to fill the array. In the case of Rot, we should be able to assume that the array keyed with a 4-bit bitvector should have exactly 16 values.

alex-ozdemir:
> In existing FHE compilers, arrays are all sizes with power-of-2. If an
array
> is initialized with less values than a power-of-2, then the elements in
the
> array are replicated to fill the array. In the case of Rot, we should be
> able to assume that the array keyed with a 4-bit bitvector should have
> exactly 16 values.

Great! Then we can enforce that in the type system.

Another question: does rotation care about the *value type* of the array?

edwjchen:
> does rotation care about the *value type* of the array?

In the context of FHE, the values should be restricted to BitVectors. Outside of FHE, rotation shouldn't be affected by the value type.

src/ir/term/mod.rs
```diff
@@ -148,6 +148,9 @@ pub enum Op {
 
     /// Call a function (name, argument sorts, return sort)
     Call(String, Vec<Sort>, Sort),
+
+    /// Cyclic rotation of an array
```
alex-ozdemir, line 152: I'd say something about the direction of the rotation.

src/ir/term/mod.rs
```diff
@@ -1664,6 +1669,28 @@ fn eval_value(vs: &mut TermMap<Value>, h: &FxHashMap<String, Value>, c: Term) ->
...
+        Op::Rot(i) => {
+            let a = vs.get(&c.cs[0]).unwrap().as_array().clone();
+            let iter = match check(&c.cs[0]) {
+                Sort::Array(k, _, s) => (*k).clone().elems_iter_values().take(s).enumerate(),
+                _ => panic!("Input type should be Array"),
+            };
+            let (mut res, len) = match check(&c) {
+                Sort::Array(k, v, n) => (Array::default((*k).clone(), &v, n), n),
+                _ => panic!("Output type of rot should be Array"),
+            };
```
alex-ozdemir, line 1681: Can you get this from the input type?

alex-ozdemir (changes requested):
Almost there!

src/ir/term/mod.rs
edwjchen: Should I define a left and right rotation, e.g. `RotLeft(usize), RotRight(usize)` if we're using `usize` for the rotation parameter? Or should I change the parameter to `i32` and extend just the frontend?

src/ir/term/mod.rs
edwjchen: By input type, do you mean the first argument instead of `c`?

src/ir/term/mod.rs
alex-ozdemir: Your call. Define whatever is needed for the FHE backend.

Just write it down :)

src/ir/term/mod.rs
alex-ozdemir: Yes. i.e., in the above match.

alex-ozdemir (approved):

alex-ozdemir:
Nice work!

edwjchen:
> Nice work!

woo!

PR #113, opened 2022-10-18
Function Abstractions

edwjchen:
Initial commit for support function abstractions.

`Computations` are maps of function names `str` to computation `Computation`. 

Updates:
- Frontend `Gen` returns `Computations` with only a single main `Computation`.
- Optimizations take in `Computations` instead of a single `Computation`. Optimizations are applied to each `Computation`. 
- Backends take in only the main `Computation`.

### DIFF
--- a/scripts/build_mpc_c_test.zsh
+++ b/scripts/build_mpc_c_test.zsh
@@ -30,89 +30,89 @@ function mpc_test_2  {
     RUST_BACKTRACE=1 measure_time $BIN --parties $parties $cpath mpc --cost-model "hycc" --selection-scheme "a+b"
 }
 
-# mpc_test_2 2 ./examples/C/mpc/playground.c
+mpc_test_2 2 ./examples/C/mpc/playground.c
 
-# build mpc arithmetic tests
-mpc_test 2 ./examples/C/mpc/unit_tests/arithmetic_tests/2pc_add.c
-mpc_test 2 ./examples/C/mpc/unit_tests/arithmetic_tests/2pc_sub.c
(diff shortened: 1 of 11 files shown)


src/ir/term/mod.rs
```diff
@@ -2093,5 +2093,29 @@ impl Computation {
     }
 }
 
+#[derive(Clone, Debug, Default)]
+/// A map of IR computations.
+pub struct Computations {
+    /// A map of function name --> function computation
+    pub cs: BTreeMap<String, Computation>,
```
alex-ozdemir, line 2100: Can you document the reason for using a treemap instead of a hashmap.

alex-ozdemir: Also, Let's use a longer name for this field. Especially since `cs` means "children" elsewhere.

scripts/build_mpc_c_test.zsh
```diff
@@ -30,6 +30,7 @@ function mpc_test_2  {
     RUST_BACKTRACE=1 measure_time $BIN --parties $parties $cpath mpc --cost-model "hycc" --selection-scheme "a+b"
 }
 
+
```
alex-ozdemir, line 33: I'd remove this change.

alex-ozdemir (changes requested):
Close! Thanks for doing this :)

src/ir/term/mod.rs
edwjchen: > Can you document the reason for using a treemap instead of a hashmap.

No reason in particular, changed to FxHashMap!

alex-ozdemir (approved):
Nice work!

PR #114, opened 2022-10-19
Function Abstraction, end-to-end testing with MPC backend

edwjchen:
Updated MPC unit tests to use function abstractions. Most of these updates are made to the C frontend and MPC lowering pass. Additional optimization and ILP solving passes need to be updated.

_Some C benchmarks timeout on ILP solving:_
- The assignment and cost model passes now need to factor in Array and Tuple semantics. This requires changes to the CostModel assignments. This also requires the the def-use-graph implementation for constant-index propagation. 

Current work-around: 
- disabled HyCC benchmark tests
- disabled tests that use Op::Select / Op::Store


scripts/aby_tests/test_inputs/sub_2.txt
```diff
@@ -1,3 +1,3 @@
 a 2 
 b 3
-res 4294967295
\ No newline at end of file
+res -1
```
alex-ozdemir, line 3: :tada:

scripts/build_mpc_c_test.zsh
```diff
@@ -110,9 +110,9 @@ mpc_test 2 ./examples/C/mpc/unit_tests/misc_tests/2pc_millionaires.c
 mpc_test 2 ./examples/C/mpc/unit_tests/misc_tests/2pc_multi_var.c
 
 # build hycc benchmarks
-mpc_test 2 ./examples/C/mpc/benchmarks/biomatch/2pc_biomatch.c
-mpc_test 2 ./examples/C/mpc/benchmarks/kmeans/2pc_kmeans.c
-mpc_test 2 ./examples/C/mpc/benchmarks/kmeans/2pc_kmeans_og.c
+# mpc_test 2 ./examples/C/mpc/benchmarks/biomatch/2pc_biomatch.c
+# mpc_test 2 ./examples/C/mpc/benchmarks/kmeans/2pc_kmeans.c
+# mpc_test 2 ./examples/C/mpc/benchmarks/kmeans/2pc_kmeans_og.c
```
alex-ozdemir, line 115: Just to confirm: Why you made this change?

alex-ozdemir (approved):
One more comment: did you change the permissions of `driver.py` intentionally? If not, I'd revert the change.

Anyways, this looks good to me. I didn't look too closely at the C front-end (trusting you here) and I ignored the ABY backend (I don't know much about it anyway). I did look closely at everythign else, and it is :heavy_check_mark: .

edwjchen:
> One more comment: did you change the permissions of `driver.py` intentionally? If not, I'd revert the change.
> 
> Anyways, this looks good to me. I didn't look too closely at the C front-end (trusting you here) and I ignored the ABY backend (I don't know much about it anyway). I did look closely at everythign else, and it is heavy_check_mark .

Reverted `driver.py`! I think the permissions changed when I edited the file from GitHub web, not sure why it does that...

scripts/build_mpc_c_test.zsh
edwjchen: Made this change because 0 cost operations (from Select and Store) in the ILP solver will sometimes cause an infinite loop. 

I'll adjust the commit message, panic on Select and Store in the ILP solver, and comment out all Array tests. I'll uncomment these tests when I merge in the def-use-graph changes.

src/target/aby/assignment/ilp.rs
```diff
@@ -77,14 +77,27 @@ fn build_ilp(c: &Computation, costs: &CostModel) -> SharingMap {
...
                 }
             }
+            Op::Select | Op::Store => {
+                for ty in &[ShareType::Boolean, ShareType::Yao] {
+                    let name = format!("t_{}_{}", i, ty.char());
+                    let v = ilp.new_variable(variable().binary(), name.clone());
+                    term_vars.insert((t.clone(), *ty), (v, 0.0, name));
+                    vars.push(v);
+                }
+            }
```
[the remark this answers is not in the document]
edwjchen: That sounds like a better approach! This was the older version of the code before adding in the def-use-graphs to propogate the constant indices. I'll panic instead and update the commit when I merge in def-use-graphs.

PR #121, opened 2022-11-17
bump deps to avoid yanked packages ; fix some clippy lints in circ_fields

kwantam:
@emlaufer @alex-ozdemir I think this fixes some recent issues with builds. Let's see what CI thinks of my thinking.


alex-ozdemir:
Great! Thanks Riad (and Evan!).

PR #126, opened 2022-12-22
add configuration system, rm DFL_T 

alex-ozdemir:
* add circ::cfg::CircCfg that is meant to be passed everywhere
* it's constructible from circ_opt::CircOpt
  * implements clap::Args, so you can set it from your compiler's
    CLI/envvars
  * defined in external crate to keep clap out of our main build
  * organized by circ module, but **not** feature gated
    * no point: the build wouldn't meaningfully change
  * includes a way to set the default field
* killed DFL_T by threading the default field around
* now, only statics are caches


alex-ozdemir:
Closing this for now. Upon a little reflection, I think it would be better to store the configuration statically.

PR #152, opened 2023-03-15
Linker Optimization Pass

edwjchen:


src/ir/term/mod.rs
```diff
@@ -2002,6 +2002,23 @@ impl ComputationMetadata {
     pub fn remove_var(&mut self, name: &str) {
         self.vars.remove(name);
     }
+
+    /// Create a call term, given the input arguments in a fixed order.
```
alex-ozdemir, line 2006: I'd describe the parameters here.

PR #163, opened 2023-06-14
r1cs: fix boolean majority

alex-ozdemir:
for m = majority(a,b,c),
we were lowering

    (-ab + bc + ac - 2abc)

instead of

    (ab + bc + ac - 2abc)

Whoops! This is a functional correctness bug: we were lowering soundly and completely, for the wrong specification.

Credit to Anna for raising the issue!

### DIFF
--- a/src/target/r1cs/trans.rs
+++ b/src/target/r1cs/trans.rs
@@ -10,7 +10,7 @@ use crate::target::r1cs::*;
 
 use circ_fields::FieldT;
 use circ_opt::FieldDivByZero;
-use log::debug;
+use log::{debug, trace};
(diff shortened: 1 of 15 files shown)


src/target/r1cs/trans.rs
```diff
@@ -528,7 +529,7 @@ impl<'cfg> ToR1cs<'cfg> {
                     //   where i = ab
                     // m = i + c(b + a - 2i)
                     let i = self.mul(a.clone(), b.clone());
-                    self.mul(c, b + &a - &(i.clone() * 2)) - &i
```
alex-ozdemir, line 531: This is the fix.

PR #182, opened 2024-01-03
optimize bellman ZK prover backend

alex-ozdemir:
three changes:
* faster integer->ff conversion
* parallel construction of bellman LCs
* parallel R1CS checking (on the CirC side)

The first change is the most important, by far. Our previous integer->ff conversion was very slow.


alex-ozdemir:
A profile before:
profile-before

And one after:
profile-after

(You may need to open these in a new tab).

The big difference is that circuit synthesis time essentially goes to zero.
Now, the obvious remaining bottleneck is deserialization.

Issue #208, opened 2024-09-27
Support for Zokrates braces syntax


StefanosChaliasos:
@cmlsharp @alex-ozdemir, how complicated would adding support for Zokrates with braces syntax be?

Most of the Zokrates programs using that syntax (e.g., https://zokrates.github.io/examples/rng_tutorial.html)

alex-ozdemir:
If the difference is purely one of concrete syntax, I don't think it would be very hard.

Perhaps you could get away with just having two different pest grammars?

If there are differences in the abstract syntax or the semantics, then you'd have to fork the Z# front-end and modify it. The modifications would probably be small, but maintaining two forks would have some downsides.

StefanosChaliasos:
Hm, there are some differences in the abstract tree. I have made most of the required changes. One question I have is that in the existing Z# implementation you are using an extension of Zokrates, right? In any case, how should I proceed? I guess for now, we can keep it as a distinct branch. Or should I create a PR that will include two distinct Zokrates front-ends? Maye zsharp and zsharpcurly? At the moment, the implementation is pretty hacky and ugly. I will need to clean it a bit if we want to merge it.

alex-ozdemir:
> Hm, there are some differences in the abstract tree.

Ah, too bad! Well, that's not a show stopper.

> I guess for now, we can keep it as a distinct branch. Or should I create a PR that will include two distinct Zokrates front-ends? Maybe zsharp and zsharpcurly?

If you're down to the work, I like the second option and I am very than happy to do the code-reviewing. If not, no worries---I understand.

StefanosChaliasos:
Sounds good, will try to create a PR :-) Your feedback will be much appreciated!

PR #211, opened 2024-10-30
Add support for ZoKrates with curly brackets (wip)

StefanosChaliasos:
This PR tries to address #208. 
Some of the features that are not supported by the grammar are:
* Integer
* cond_store
* Array Metadata
* Witness

Furthermore, we don't currently support MPC.

Finally, the implementation of assembly has not been tested and shouldn't be complete. 

All the tests pass. To run them, you can use the following commands:

```
cargo build --release --features=r1cs --features=smt --features=zok --features=lp --features=bellman --features=zokc --example=circ
cargo build --release --features=r1cs --features=smt --features=zokc --features=lp  --features=bellman --example=zcxi
./scripts/zokrates_curly_test.zsh
```

The changes to support ZoKrates with curly brackets are the following:

* We added `third_party/ZoKratesCurly`.
* We added a new frontend, `src/front/zsharpcurly`.
* We added a new feature called `zokc`.
* We updated `circ` executable to support zsharpcurly and implemented `zcxi`.
* Finally, we copied and updated the tests and the scripts to run the tests for zsharpcurly.

TODOs:
- [x] Check that all tests have been converted correctly (which are not!)
- [x] Fix assembly
- [x] Apply the proper format
- [x] Do we want to support MPC?

Note that you can check the changes vs the original implementation in the first 4 commits of this PR.


src/front/zsharpcurly/interp.rs
```diff
@@ -45,5 +45,10 @@ pub fn extract(
                 })
                 .collect::<Result<Vec<_>, _>>()?,
         )),
+        Ty::Tuple(tys) => Ok(T::new_tuple(
+            tys.iter()
+                .map(|ty| extract(name, ty, scalar_input_values))
```
alex-ozdemir, line 50: I would expect the `name` argument to the recursive call to be different for each iteration. Why is that not the case?

src/front/zsharpcurly/mod.rs
```diff
@@ -957,37 +1035,21 @@ impl<'ast> ZGen<'ast> {
...
+        let ret_ty = f.return_type.map(|r| self.type_(&r));
         // set up stack frame for entry function
         self.circ_enter_fn(n.to_owned(), ret_ty.clone());
-        let mut persistent_arrays: Vec<String> = Vec::new();
+        let persistent_arrays: Vec<String> = Vec::new();
         for p in f.parameters.iter() {
             let ty = self.type_(&p.ty);
             debug!("Entry param: {}: {}", p.id.value, ty);
-            let md = self.interpret_array_md(&p.array_metadata);
+            // XXX(unimpl) array metadata
```
alex-ozdemir, line 1049: Would it be better to interpret the array metadata and panic if it is non-empty? Silently ignoring it seems like a bad idea.

src/front/zsharpcurly/mod.rs
```diff
@@ -1074,55 +1129,29 @@ impl<'ast> ZGen<'ast> {
...
-                                "Party number {num_val} greater than the number of parties ({n_parties})"
-                            ),
-                            &private.span,
-                        )
-                    }
+            Some(ast::Visibility::Private(_)) => match self.mode {
+                Mode::Proof | Mode::Opt | Mode::ProofOfHighValue(_) => ZVis::Private(PROVER_ID),
+                Mode::Mpc(_n_parties) => {
+                    // XXX(unimpl) party number
+                    panic!("Mpc mode is not implemented");
```
alex-ozdemir, line 1154: I'd add that this is for zsharpcurly

src/front/zsharpcurly/mod.rs
```diff
@@ -1814,22 +1876,25 @@ impl<'ast> ZGen<'ast> {
             );
         }
 
-        if let Some(ast::ArrayParamMetadata::Transcript(_)) = &c.array_metadata {
-            if !value.type_().is_array() {
-                self.err(format!("Non-array transcript {}", &c.id.value), &c.span);
-            }
-            self.mark_array_as_transcript(&c.id.value, value.clone());
-        }
+        // if let Some(ast::ArrayParamMetadata::Transcript(_)) = &c.array_metadata {
```
alex-ozdemir, line 1879: Again, I suggest throwing an error.

alex-ozdemir (changes requested):
This is a very nice patch, Stefanos. I have a few questions about it, but I think they should be easy to address.

src/front/zsharpcurly/mod.rs
StefanosChaliasos: In the [new grammar](https://github.com/StefanosChaliasos/circ/blob/zsharpcurly/third_party/ZoKratesCurly/zokrates_parser/src/zokrates.pest), metadata is not supported, so I guess we should skip it. Should I remove this comment from here?

src/front/zsharpcurly/mod.rs
StefanosChaliasos: Fixed

src/front/zsharpcurly/mod.rs
StefanosChaliasos: I removed the commented-out code from here. Basically, the new AST does not contain ArrayParamMetadata

src/front/zsharpcurly/interp.rs
StefanosChaliasos: Nice catch, tuples were not implemented correctly. I updated the code in a few positions and added test cases. Here, we do something similar to arrays.

src/front/zsharpcurly/mod.rs
alex-ozdemir: Ah! I understand now. Yes, having metadata give a parsing error is fine. You can remove the comment or not, whichever you prefer.

alex-ozdemir (approved):
This looks good to me now.
If you're happy with it, then I'll merge it.

src/front/zsharpcurly/mod.rs
```diff
@@ -0,0 +1,2381 @@
...
+        if IS_CNST {
+            self.cvar_assign(name, new)
+        } else {
+            self.circ_assign(Loc::local(name.to_string()), Val::Term(new))
+                .map_err(|e| format!("{e}"))
+                .map(|_| ())
+        }
+    }
+
+    fn assembly_assign_impl_<const IS_CNST: bool>(
```
StefanosChaliasos, line 585: @alex-ozdemir, can you review this code?

src/front/zsharpcurly/mod.rs
```diff
@@ -0,0 +1,2381 @@
...
+            let new = loc_store(old, &zaccs[..], new_val).map(const_fold)?;
+
+            debug!("Assembly Witness Assign: {}", name);
+            self.circ_assign(Loc::local(name.to_string()), Val::Term(new))
+                .map_err(|e| format!("{e}"))
+                .map(|_| ())
+        }
+    }
+
+    fn assembly_constraint_<const IS_CNST: bool>(
```
StefanosChaliasos, line 634: Here I just add the assembly constraints like it was an assertion.

StefanosChaliasos:
I think that if we make sure that assembly is handled correctly, then we can proceed. ref: https://zokrates.github.io/language/assembly.html

alex-ozdemir:
Is the omission of [`field_to_bool_unsafe`](https://zokrates.github.io/language/assembly.html#field_to_bool_unsafe) intentional?

src/front/zsharpcurly/term.rs
```diff
@@ -0,0 +1,1363 @@
...
+}
+
+pub fn field_to_bits(f: T, n: usize) -> Result<T, String> {
+    match &f.ty {
+        Ty::Field => uint_to_bits(T::new(Ty::Uint(n), term![Op::PfToBv(n); f.term])),
+        u => Err(format!("Cannot do uint-to-bits on {u}")),
+    }
+}
+
+pub fn field_to_bool_unsafe(f: T) -> Result<T, String> {
```
StefanosChaliasos, line 1113: @alex-ozdemir, can you review this one?

StefanosChaliasos:
Sorry for the delay. It was not intentional; I tried to implement it (without changing the IR), but the [test](https://github.com/circify/circ/blob/f442e2b602cabc063d0f7ccfbfba0e16deb33d5d/examples/ZoKratesCurly/pf/assembly_field_to_bool_unsafe.zok) failed.

src/front/zsharpcurly/term.rs
alex-ozdemir: I think that you actually want [this operator](https://github.com/circify/circ/blob/master/src/ir/term/mod.rs#L201) using it will maintain the connection between the input and the output of the operator in the final constraint system, it will just assume that the value is 0 or 1. See [this part](https://github.com/circify/circ/blob/master/src/target/r1cs/trans.rs#L556) of R1CS lowering.

What do you think?

alex-ozdemir (changes requested):
(A change may not be necessary---see my comment and let me know what you think.)

src/front/zsharpcurly/term.rs
StefanosChaliasos: Thanks, that makes sense! Should we still wrap it in a new_witness Op? If we do so, in the test, should we also pass the new_witness value?

StefanosChaliasos: Fixed it

StefanosChaliasos:
@alex-ozdemir fixed the issues, now the mpc tests fail. Seems like we miss libcrypto:

```
Server error: ./third_party/bin/ABY/aby_interpreter: error while loading shared libraries: libcrypto.so.1.1: cannot open shared object file: No such file or directory
```

StefanosChaliasos:
Great, thanks! I think it's ready to be merged

alex-ozdemir:
Wonderful, it's great to have this in!

PR #219, opened 2025-02-06
Reopened: Extend input declarations in Z# to support structs

lorenzorota:
Reopened PR #218 to fix branching issues

---

Currently, program inputs cannot be declared for structs that are being returned. Take the following program
```
struct Pt {
    field x
    field y
}
struct Pts {
    Pt[2] pts
}

def main(private Pts[1] pts) -> Pt:
    return Pt {x: pts[0].pts[0].x * pts[0].pts[1].x, y: pts[0].pts[0].y * pts[0].pts[1].y}
```
with the value maps
```
(set_default_modulus 52435875175126190479447740508185965837690552500527637822603658699938581184513
(let (
    (pts.0.pts.0.x #f2)
    (pts.0.pts.0.y #f4)
    (pts.0.pts.1.x #f2)
    (pts.0.pts.1.y #f4)
) true ;ignored
)
)
```
and
```
(set_default_modulus 52435875175126190479447740508185965837690552500527637822603658699938581184513
(let (
    (return.x #f4)
    (return.y #f16)
) true ;ignored
)
)
```
for the prover and verifier, respectively. This would throw a _not implemented_ error since precomputations are not handled.

This PR adds the missing precomputations, similar to how it's done for arrays and mutable arrays, i.e., computing the precomputations by unwrapping all the struct fields. For this I included two new functions in `T`: `unwrap_struct` and `unwrap_struct_ir`, which both return a FieldList or an error.


alex-ozdemir (approved):
This looks great. Once the build passes, I'll merge it.
I recommend you rebase/merge in the new `master` and then fix any remaining build issues.

lorenzorota:
I'm running the lint checker locally and I get the following error:
```
error[E0432]: unresolved import `std::num::NonZero`
 --> circ_hc/src/collections/lru.rs:3:5
  |
3 | use std::num::NonZero;
  |     ^^^^^^^^^^-------
  |     |         |
  |     |         help: a similar name exists in the module: `NonZeroI8`
  |     no `NonZero` in `num`
  ```
I'll see if this has something to do with my version of rust. Does this also happen on your end?

lorenzorota:
> I'll see if this has something to do with my version of rust. Does this also happen on your end?

Fixed it now, it was indeed a rust version issue.

alex-ozdemir:
Great, thanks for this patch!

PR #232, opened 2025-06-20
Update comment in PfChallenge to clarify how it works for multi round SNARKs

sashafrolov:


sashafrolov:
This led me on a bit of a wild goose chase until I realized this did actually work the intended way.

PR #241, opened 2026-07-20
support ROM for zsharp-curly

kmerrill18:
NOTE: Claude Opus wrote this whole code change.

Add the syntax labels needed to mark arrays as RAM or ROM using the ram extension in CirC to ZoKrates Curly, the new version of ZoKrates. Actually pipe through the ROM marking. (I didn't implement the RAM one because I don't need that one right now.)


kmerrill18:
It seems there's a flaky test as part of the CI. I got a test failure unrelated to my changes, investigated for a while, and then decided to just re-run the CI, and it passed.

